PR #3, opened 2017-12-01
add xor and xnor

steinm:


carrotIndustries:
merged with slight modifications

PR #7, opened 2018-01-05
Add STM8S003F3U6TR

fruchti:
Adds the required padstacks for the UFQFPN20 package as well.

### DIFF
--- a/entities/ic/mcu/stm/stm8s003f3.json
+++ b/entities/ic/mcu/stm/stm8s003f3.json
@@ -1,0 +1,19 @@
+{
+    "gates": {
+        "37d6ce5e-d5d4-4aa7-bb90-ebeca29453c5": {
+            "name": "",
+            "suffix": "",
+            "swap_group": 0,
+            "unit": "bc8b241e-8de2-4551-9670-5869edafb70f"
+        }
+    },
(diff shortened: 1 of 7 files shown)


carrotIndustries:
Looks got to me overall, some minor nitpicks:
 - The NRST pin is low active, so add the inverted decoration to the symbol pin
 - The package is missing the parameter program for adjusting the courtyard

fruchti:
Thanks for the feedback. I added the courtyard parameter program and moved the padstacks to the package directory. I did not add the inverted decoration, because that would be a double inversion in my opinion (it seems ST consistently uses a capital N for active-low pins).

fruchti:
Fixed the folder name typo as well. Wasn't aware this was as easy as moving files and deleting pool.db.

carrotIndustries:
In fact, you don't even have to delete the `pool.db` just click 'update pool' in the pool manager to rebuild the database. The courtyard still isn't quite right: https://github.com/carrotIndustries/horizon/wiki/Creating-a-Package#courtyard 

>  At 0mm courtyard expansion the courtyard outline is the (rectangular) hull around copper pads and package outline.

fruchti:
Thanks! Should be all fixed now.

PR #11, opened 2018-01-10
 Add STM32F103C8T6TR and STM32F103CBT6TR 

fruchti:


carrotIndustries:
LGTM overall, some minor nitpicks:
 - It'd be nicer to have power supply pins at the top/bottom edge
 - One pin in the symbol can be connected to more than one pad, so only add one VDD pin in the unit and map all required pads to this pin
 - silkscreen clearance is a bit marginal, 0.2mm clearance to exposed copper seems like a reasonable clearance.

fruchti:
I extended the silkscreen clearance to both package outline and copper to 0.2 mm. Looks best when it's symmetric IMHO.

I combined the power pins, but did not shift them to top and bottom of the symbol for a number of optical reasons:
- The pin names of vertical pins always get in the way of other pin names inside the symbol outline and I have yet to find a part where this looks pretty. As far as I see, Horizon doesn't have the capabilities to rotate pin names, so the symbol would have to be much larger just so the pin names could “dodge” each other.
- In my opinion, the placement of, especially if there are multiple, bypass capacitors looks much cleaner if the power pin's on the side of the symbol.
- For microcontrollers, I like all the supplementary passive components to be on the left side, so the right one is free for all I/O. That's why NRST and the pins for the crystals are on the left side in my symbol as well.

Of course, all of this is personal preference and I don't have a problem with leaving this pull request open until there exists a clear convention on where supply pins should go in every case.

carrotIndustries:
Regarding the symbol: The only "guide" on how schematic symbols I've found so far is the kicad library convention which specifies power pins to be top/bottom: http://kicad-pcb.org/libraries/klc/S4.2/ Are you aware of any recognized standard documents covering this? The only ones I found so far deal with the basic stuff like diodes and capacitors but don't give any recommendations on how to draw nonstandard symbols. I'd like to avoid coming up with a new convention for the horizon pool if there are existing conventions to use.

fruchti:
I didn't find anything apart from the KLC either. I think it would would be pretty hard to have a general convention which would look good and lead to neat schematics in every case. A convention is clearly needed, but I don't know if a rule anymore concrete than the KLC's “group similar pins together” is really beneficial because it will always look odd with some types of parts and thus lead to ambiguity.

Maybe a more abstract, yet measurable rule like “The pin configuration, where the part's recommended configuration with all supporting passives needs the least space on the schematic” would be an option, too.

A few days ago, I wanted to set up a library for simple pin headers and I noticed unsure I was without a convention. I'd prefer to set up a rough convention and refine it whenever some ambiguity is noticed, maybe with a poll on the different options. I could start a rough draft based on the KLC and IPC footprint rules in the next few days because that would help a lot in getting the process going.

fruchti:
Don't get me wrong, I absolutely think that the established conventions are good in that they make mostly sensible decisions and because people are used to them, of course. But when designing a all-new library there's a huge potential of rethinking the design assumptions for the convention, because there are no legacy parts which have to somehow fit into the convention. I like to keep that in mind and rather discuss a bit longer on the convention than having to retro-fit some rules which turn out to be non-optimal.

carrotIndustries:
> I could start a rough draft based on the KLC and IPC footprint rules in the next few days because that would help a lot in getting the process going.

Go ahead! That's what I thought of in https://github.com/carrotIndustries/horizon-pool/issues/8 as well. Having a clear convention in place makes it much easier for everyone to contribute parts to the pool.

Having power/ground on top/bottom is the way I've always done it and seen it done in most places. Being able to rotate pin names shouldn't be that hard to implement, let's see..

fruchti:
Another way would be to almost always split the power pins into a separate unit. The only exception would be parts where the supply pins fulfill more of an analog purpose, like switching regulators or supply monitor ICs.

In this particular case the controller would have two supply units for the digital and the analog supply. That would be nice, too, because you can put a inductor or ferrite bead or whatever you like neatly between them.

carrotIndustries:
IMO, splitting out the power supply pins doesn't make  much sense for most devices with much less than 100 pins since it leads to more complex schematics.

fruchti:
I rebased this PR since it wasn't able to merge without conflicts any more.

pwnorbitals:
This would be useful, is merging still an option @carrotIndustries ? Thanks :)

rroohhh:
Bot!

carrotIndustries:
- Power pins should go on top/bottom
 - Symbol orientations don't really make sense for large symbols like this one

fruchti:
Bot!

fruchti:
> * Power pins should go on top/bottom

I’ll add this to the convention as well (and to https://github.com/horizon-eda/horizon-pool-convention/issues/7). Unless you have objections, I would also add a rule to prefer the ‘perpendicular’ name orientation unless this makes the symbol unreasonably wide.

Fixed everything the bot complained about and redid the symbol. I don’t really know why the multiple symbol orientations popped up since I didn’t see the as set in the symbol editor. Let me know if you have additional comments!

fruchti:
Bot!

PR #17, opened 2018-01-13
Equiped resistors with silkscreen of .15mm and rounded pad option

greenscreenflicker:
Equiped resistors with silkscreen of .15mm and rounded pad option

### DIFF
--- a/packages/passive/smd/R0402-rnd/package.json
+++ b/packages/passive/smd/R0402-rnd/package.json
@@ -1,6 +1,6 @@
 {
     "_imp": {
-        "grid_spacing": 100000,
+        "grid_spacing": 1250000,
         "layer_display": {
             "layer_opacity": 50.0,
             "layers": {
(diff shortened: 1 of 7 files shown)


greenscreenflicker:
@carrotIndustries can you check?

carrotIndustries:
Looks good overall, but split the two unrelated changes (new packages and changed packages) into two PRs.

some notes on the new packages:
 - set these as an alternate package for their counterparts with non-rounded pads, so they can easily be used instead of them
 - the parameter corner radius is meant to be applied to pads, setting it in the package's parameters has no effect
 - package names aren't bound to unix/dos filename conventions, so feel free to name them like "C0603 (rounded pads)"

greenscreenflicker:
One comment on the alternative packages: This created a kinda crash, as mentioned here:
https://github.com/carrotIndustries/horizon/issues/62
Since then, I trying not to use alternative packages.

E.g. if you set both packages as alternative, issue 62 happens also. So maybe, if you can fix that?

PR #19, opened 2018-01-17
Added opamp1

greenscreenflicker:
Added unit, entry, symbol and a sample part (MCP6001T) for OpAmp1

### DIFF
--- a/entities/linear/opamp/opamp1/opamp1.json
+++ b/entities/linear/opamp/opamp1/opamp1.json
@@ -1,0 +1,16 @@
+{
+    "gates": {
+        "742b88b4-c09c-47f4-a0af-3498b5b7b7ce": {
+            "name": "Opamp",
+            "suffix": "",
+            "swap_group": 1,
+            "unit": "2c88d43f-5d7f-4a77-b983-19506649fedd"
+        }
+    },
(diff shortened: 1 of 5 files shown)


carrotIndustries:
There are several things wrong with the opamp symbol/unit/entity:
 - don't integrate power supply pins, instead reference the existing unit 'single power in the entity', this makes the unit usable for dual/quad opamps
 - pin names should be uppercase
 - use more descriptive names, see the logic gates
 - using the 'inverted' pin decoration on an opamp is ... unusual, instead show the pin names or draw a +/- using lines
 - leave manufacturer field blank for generic units, same goes for the entity (not entry)
 - symbol is too tiny, double the spacing between the two input pins and scale the rest accordingly
 - in entities with a single gate, this gate is usually called "Main"

Package:
 - assembly outline is wrong, see other packages on how the assembly outline is supposed to look like
 - assembly outline is missing pin 1 indicator
 - assembly refdes text has nonzero width
 - 3d models path uses backslashes for path separator (my bad, need to replace backslashes automatically)

Part:
 - you didn't specify the complete MPN, "MCP6001T-I/OT" would be an example for a proper MPN
 - value has to be empty for ICs
 - datasheet field has to be an actual link
 - tags are missing

PR #20, opened 2018-01-18
Added MCP6001T-I/OT

greenscreenflicker:
Added first operational amplifier, based on feedback of https://github.com/carrotIndustries/horizon-pool/pull/19

I tried to incoperate every feedback, i hope i considered everything, please check again and provide feedback.

If that fits, i add opamp2 and opamp4.

### DIFF
--- a/entities/linear/opamp/opamp1/opamp1.json
+++ b/entities/linear/opamp/opamp1/opamp1.json
@@ -1,0 +1,25 @@
+{
+    "gates": {
+        "45e69187-ca20-444e-a041-3f5f995b9e0b": {
+            "name": "Power",
+            "suffix": "Z",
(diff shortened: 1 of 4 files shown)


carrotIndustries:
This PR is missing a symbol for the new unit.

carrotIndustries:
- Symbol is missing
 - Power Gate is named 'Z'. What's going on?
 - Entity still has 'generic' manufacturer
 - Silkscreen will be below package outline
 - There are superflous junctions in the package
 - Names are still wrong
 - Don't put the generic part number in tags

greenscreenflicker:
@carrotIndustries  Silkscreen is not below the package, checked with 3d model.

PR #22, opened 2018-01-18
Added round 0402,0603,0805,1206 smd footprint

greenscreenflicker:
Added round 0402,0603,0805,1206 smd footprint as requested in https://github.com/carrotIndustries/horizon-pool/pull/17


carrotIndustries:
Looks good overall, but the R0805's assembly refdes is positioned incorrectly.

PR #23, opened 2018-01-23
Added MCP6001T-I/OT #2

greenscreenflicker:
Tryed to incoperate every issue from https://github.com/carrotIndustries/horizon-pool/pull/20

Known issue: 3D model missing due to current software issue.

### DIFF
--- a/entities/linear/opamp/opamp1/opamp1.json
+++ b/entities/linear/opamp/opamp1/opamp1.json
@@ -1,0 +1,24 @@
+{
+    "gates": {
+        "45e69187-ca20-444e-a041-3f5f995b9e0b": {
+            "name": "Power",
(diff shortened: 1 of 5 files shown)


greenscreenflicker:
@carrotIndustries ping

carrotIndustries:
- Symbol is missing $REFDES/$VALUE text...
 - You tripled instead of doubled the opamp pin spacing
 - Make sure that the symbol nicely lines up with the "single power symbol"
 - Package still has superfluous junctions
 - tags should be lower case (see other tags)
 - even though the silkscreen outline may fit the 3D model, the datasheet of this particular part for example doesn't specify how much the packages shrinks near the bottom. Place the silkscreen all outside of the package outline.

PR #24, opened 2018-01-29
Updated smd footprints with silkscreen of .15mm

greenscreenflicker:
fixed https://github.com/carrotIndustries/horizon-pool/pull/21


greenscreenflicker:
@carrotIndustries ping.

carrotIndustries:
In R0603, the 3d model filename uses backslashes instead of slashes.

fruchti:

PR #25, opened 2018-02-01
Added footprints for PDIP (plastic dual inline package)

atoav:
Added footprints for PDIP (Plastic Dual-Inline-Packages) for pin counts 6 to 32. The pad and hole sizes are passed on Level B of IPC-2222¹ with the assumption of 0.55mm maximum lead diameter (which fits both my own measurements of various IC legs, precision and non-precision IC-sockets as well the data-sheets of multiple IC-sockets I checked. 

The DIP package is a bit old-school, but is still used widespread in DIY kits and the maker scene, because it is easier to handsolder than SMD devices.

¹ see: [https://www.pcb-3d.com/knowledge-base/pth-dimensions/](https://www.pcb-3d.com/knowledge-base/pth-dimensions/)

### DIFF
--- a/packages/ic/th/dip/direct_solder/DIP10_7.62mm_lead_span_2.54mm_pitch/package.json
+++ b/packages/ic/th/dip/direct_solder/DIP10_7.62mm_lead_span_2.54mm_pitch/package.json
@@ -1,0 +1,642 @@
+{
+    "_imp": {
+        "grid_spacing": 1250000,
+        "layer_display": {
+            "layer_opacity": 50.0,
+            "layers": {
+                "-1": {
+                    "color": {
+                        "b": 0.0,
+                        "g": 1.0,
+                        "r": 1.0
+                    },
+                    "display_mode": "fill",
+                    "visible": false
+                },
+                "-100": {
+                    "color": {
+                        "b": 0.0,
+                        "g": 0.5,
+                        "r": 0.0
+                    },
+                    "display_mode": "fill_only",
+                    "visible": true
+                },
+                "-110": {
+                    "color": {
+                        "b": 0.25,
+                        "g": 0.5,
+                        "r": 0.25
+                    },
+                    "display_mode": "outline",
+                    "visible": true
+                },
+                "-120": {
+                    "color": {
+                        "b": 0.8999999761581421,
+                        "g": 0.8999999761581421,
+                        "r": 0.8999999761581421
+                    },
+                    "display_mode": "fill_only",
+                    "visible": false
+                },
+                "-130": {
+                    "color": {
+                        "b": 0.800000011920929,
+                        "g": 0.800000011920929,
+                        "r": 0.800000011920929
+                    },
+                    "display_mode": "fill",
+                    "visible": false
+                },
+                "-140": {
+                    "color": {
+                        "b": 0.5,
+                        "g": 0.5,
+                        "r": 0.5
+                    },
+                    "display_mode": "outline",
+                    "visible": true
+                },
+                "-150": {
+                    "color": {
+                        "b": 0.5,
+                        "g": 0.5,
+                        "r": 0.5
+                    },
+                    "display_mode": "outline",
+                    "visible": true
+                },
+                "-160": {
+                    "color": {
+                        "b": 0.5,
+                        "g": 0.5,
+                        "r": 0.5
+                    },
+                    "display_mode": "outline",
+                    "visible": true
+                },
+                "0": {
+                    "color": {
+                        "b": 0.0,
+                        "g": 0.0,
+                        "r": 1.0
+                    },
+                    "display_mode": "fill",
+                    "visible": true
+                },
+                "10": {
+                    "color": {
+                        "b": 0.5,
+                        "g": 0.5,
+                        "r": 1.0
+                    },
+                    "display_mode": "outline",
+                    "visible": true
+                },
+                "20": {
+                    "color": {
+                        "b": 0.8999999761581421,
+                        "g": 0.8999999761581421,
+                        "r": 0.8999999761581421
+                    },
+                    "display_mode": "fill_only",
+                    "visible": true
+                },
+                "30": {
+                    "color": {
+                        "b": 0.800000011920929,
+                        "g": 0.800000011920929,
+                        "r": 0.800000011920929
+                    },
+                    "display_mode": "fill",
+                    "visible": false
+                },
+                "40": {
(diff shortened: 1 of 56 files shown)


carrotIndustries:
The copper side of things looks fine, the other layers still need some work:
 - Courtyard parameter program is missing
 - Package/Assembly should indicate the outline of the plastic part
 - The silkscreen text should be visible after assembly

Still not quite sure on how to design the silkscreen for DIP parts:
 - The silkscreen in your packages isn't clearly visible after assembly
 - I ran the "IPC library expert" to get an idea of the IPC recommendations: This one outputs something similar to the SMD ones (elongated line on the short side of the package with pin 1 on it)
 - For TH packages, a dot/triangle may be appropriate as well

@fruchti  any more ideas?

atoav:
How do I add a Courtyard parameter program? If I missed it in the wiki – sorry for asking.

I had a hard time to indicate the plastic part/assembly, because there is no real generic shape there. You don't really know whether people are going to use a IC socket or solder an IC directly. The IC-socket will be much bigger than the IC. I wasn't sure if it is a good idea to make seperate packages for both variants (which I could definitly do if so required). 

The silkscreens were also made with IC sockets in mind, but I can definitly try to orient myself on other stuff. I work a lot with people/beginners who DIY stuff and the mnemoic silkscreen with the semicircle orientation definitly helps people to assemble stuff right.

fruchti:
With the IPC rules, I imagine something similar to SO packages for the silkscreen, too. I don't think an additional pin 1 indicator outside of the package (where it would be visible after assembly) makes sense because it would overlap with pads of other packages in a lot of cases. For connectors, such a pin 1 indicator would be good, though (think pin headers).

Personally, I used the “notch”-style for all my DIP packages as well, because it's so intuitively clear. That's pretty much a tradeoff between IPC-compliance and hobbyist-friendlyness and we will have to adress that sooner or later. In my opinion, the rule that all silkscreen has to be visible after assembly might make some economic sense, but isn't really helpful for prototyping and small-scale production.

I'd vote vor separate packages for DIPs with and without socket as they're pretty different after all.

fruchti:
With the IPC rules, I imagine something similar to SO packages for the silkscreen, too. I don't think an additional pin 1 indicator outside of the package (where it would be visible after assembly) makes sense because it would overlap with pads of other packages in a lot of cases. For connectors, such a pin 1 indicator would be good, though (think pin headers).

Personally, I used the “notch”-style for all my DIP packages as well, because it's so intuitively clear. That's pretty much a tradeoff between IPC-compliance and hobbyist-friendlyness and we will have to adress that sooner or later. In my opinion, the rule that all silkscreen has to be visible after assembly might make some economic sense, but isn't really helpful for prototyping and small-scale production.

I'd vote vor separate packages for DIPs with and without socket as they're pretty different after all.

fruchti:
With the IPC rules, I imagine something similar to SO packages for the silkscreen, too. I don't think an additional pin 1 indicator outside of the package (where it would be visible after assembly) makes sense because it would overlap with pads of other packages in a lot of cases. For connectors, such a pin 1 indicator would be good, though (think pin headers).

Personally, I used the “notch”-style for all my DIP packages as well, because it's so intuitively clear. That's pretty much a tradeoff between IPC-compliance and hobbyist-friendlyness and we will have to adress that sooner or later. In my opinion, the rule that all silkscreen has to be visible after assembly might make some economic sense, but isn't really helpful for prototyping and small-scale production.

I'd vote vor separate packages for DIPs with and without socket as they're pretty different after all.

atoav:
okay @fruchti you convinced me, multiple footprints make sense here. How would we name them?

fruchti:
I thought a bit more about the silkscreen issue and came to the conclusion it's probably best to make them as IPC-compliant as possible. More protoyping-friendly versions could be later added as alternative footprints. That would be 4 footprint variants in this case but I couldn't think of a better solution.

Regarding naming: Personally, I'd prefer a more verbose style than KiCAD has. For DIPs we need pin count, lead span and pin pitch to fully specify them, right? I'd suggest names like `DIP8, 7.62 mm lead span, 2.54 mm pitch (socket)` which would translate to a file name `DIP8_7.62mm_lead_span_2.54mm_pitch_socket.json` [1]. The most important attributes would be at the front and you wouldn't have to know all the letters for the parameters by heart. Irregular DIPs with missing or fused pins wouldn't fit into this naming scheme, but they are probably better placed in the manufacturer-specific folders anyway.

@carrotIndustries Any thoughts on this?

___
[1] I don't have much free time right now but I plan writing up some short, clear rules how the file name is derived from the package name. This is roughly what I have in mind.

carrotIndustries:
> @carrotIndustries Any thoughts on this?

Looks good to me.

atoav:
@carrotIndustries What do you think about something like this:  
dip_ipc  
(Pin one indicator and indent are part of the assembly layer)

atoav:
Okay, unless somebody finds errors or there is some critique on the way I did the silkscreens, I think this one is done : )

carrotIndustries:
Having both the extended like and dot for pin 1 indicator appears somewhat redundant to me, same goes for the DIY silkscreen

 - Courtyard parameter programs are missing
 - The notches are off-center

carrotIndustries:
Looks like the last commit includes a bit more than intended: There are new unrelated 3d models and many unrelated files got CRLF line endings.

Some of the notches still are off-center.

fruchti:
One last thing to the naming: IPC-compliant footprints should be the standard and thus implicit. The same goes for the normal, direct soldering. I propose names like

 - `DIP8, 7.62 mm lead span, 2.54 mm pitch (with socket)`
 - `DIP8, 7.62 mm lead span, 2.54 mm pitch`
 - `DIP8, 7.62 mm lead span, 2.54 mm pitch (with socket, DIY silkscreen)`
 - `DIP8, 7.62 mm lead span, 2.54 mm pitch (DIY silkscreen)`

The parentheses would hold qualitative modifiers, like `SO8 […]` and `SO8 […] (manual soldering)`. Defining IPC-compliant footprints as the standard and without any text in parentheses states a clear preference for the pool and also makes adding new variants easier (otherwise, all variants would have to be renamed).

I hope you don't mind me rambling about the names of the footprints so much. Thinking about examples like this helps a lot in coming up with a general wording, at least for me it does.

atoav:
Okay the packages should be alright now, I really have no idea how I introduced these CRLF Line endings tho. Anybody with better git skills here?

PR #27, opened 2018-02-11
Added footprints for TSSOP

atoav:
Added footprints for TSSOP with pin counts from 6 to 24. Based on [this](http://cds.linear.com/docs/en/packaging/TSSOP_14_05-08-1650.pdf)


carrotIndustries:
- Courtyard parameter programs are missing
- In some packages, the assembly reference designator text is too large. U1234 should always fit within the assembly outline
 - Personally, I don't like the 'bracket' style of the bottom silkscreen likes. A straight line is sufficient.

atoav:
I am starting to implement the mentioned changes. I think having good clean packages to build uppon is very important. Do you think there exists a "sane" way to implement something like silk screen variants? As mentioned in the other pull request, I work at university with a lot of totally un-techy DIY people and there sometimes other silkscreens make sense (especially for the TH-stuff.

atoav:
I did:
- Remove Packages for TSSOP-6, TSSOP-12, TSSOP-18 and TSSOP-22 because there was no reliable data on their package sizes and they seem to be rarely used
- Verified all other footprints and package sizes
- Added Courtyard Parameter Programs
- Adjusted `$RD` sizes on the Assembly Reference Designators
- Removed Bracket and generally tried to make things more uniform

carrotIndustries:
These things still need some work:
- Some packages are named TTOP - typo?
- The -16's courtyard polygon is missing the parameter class

PR #28, opened 2018-03-05
Added TO-92 packages

atoav:
I added packages/footprints for TO-92:
- `to92-inline`
- `to92-inline_wide`
- `to92-molded`
- `to92-molded_wide`
- `to92-molded_wide_reverse`

I looked at different specifications, but endend up basing it on the [KiCAD-footprints found here](https://www.mikrocontroller.net/wikifiles/8/82/TO-92-Gehaeuse_RevB2.pdf). For the Silkscreen I didn't find any IPC style, but maybe I just looked the wrong way?


carrotIndustries:
Sorry for not getting back to you earlier, have been busy with other sutff:

 - Some packages are missing the courtyard outline
 - The courtyard should include the whole package extents
 - Many packages are missing the courtyard parameter program
 - The package outline should be a polygon
 - The assy refdes should be contained within the assy outline

carrotIndustries:
>Any Idea how to draw that arc with a polygon? Or should I just tesselate by hand?
The draw polygon allows for making an edge an arc instead of a line:
Start the edge, press 'a', then click to set the center point of the arc, then click again to set the end point. End point and start point must have the same distance from the center point, otherwise the edge will appear as a line.

carrotIndustries:
- assy refdes Q1234 won't fit
 - pin 1 indicator dot in assembly layer could be a bit bigger
 - superfluous parameters for pads (you only need pad/hole diameter)

PR #29, opened 2018-03-30
add 1x 2.54mm pitch pin headers

carrotIndustries:
This adds packages for generic single-row pin, 2.54mm pin headers
The overall design is mostly copied from the KiCad ones. I'm still not sure whether it is a good idea to have the silkscreen that close to the package body.

Any ideas @fruchti and others?


atoav:
Great, I can cross these now from my own list too. Any plans to do dual row also?  

In practice I never had any issues with similar silkscreens to the ones you used. The silkscreen for pin headers with a pin count >3 is typically only used to make clear on which side of the pcb they belong.  

So no objections from my side.

carrotIndustries:
Yeah, dual row and other variations of the theme are on my list as well. I wrote some ad-hoc C++ to generate these.

atoav:
Now that I've slept a night over it I would probably have something that could be changed:  
Although this is most of the time not an issue, it would probably be useful if male and female versions of the assembly/silkscreen could be differenciated easily.  

What you did looks already pretty much like a female pin header, so maybe we could think about adding a male version, with these tiny notches that can be seen in the top view, to make it perfectly clear.  

ebay-header-male-dimensions  

As I mentioned, in most cases (PCB stacking) it makes no big difference, but desoldering pinheaders can be a big pain and in some cases even break PCBs, thats why from a hobbyist-/DIY-standpoint it would probably be good to have them visually different (and it probably makes sense for the 3D-models too).

carrotIndustries:
The plan was to keep headers and sockets separate anyway. Using the notched outline for silkscreen looks too "noisy" IMO.

fruchti:
IMO, the silkscreen lines' centers should be either flush with the package outline or the silkscreen have the proposed 0.2 mm clearance. Having the silkscreen at +-1.27 mm would have the benefit that you could have multiple headers right next to each other (2.54 mm center-center between pins) with no silkscreen overlap, which is a great plus as far as I'm concerned. I mean, yes, your 0.075 mm offset makes sure there's no silkscreen underneath the package but it's still below our current suggestion, without any other advantages.

I'm with @carrotIndustries in that notches in the silkscreen look a bit too noisy. I don't think we need a visual distinction between male and female headers in the silkscreen.

carrotIndustries:
I'm still not entirely convinced of your silkscreen for telling sockets apart from headers since there are snap-apart sockets as well: https://www.reichelt.de/?ARTICLE=73677 

The next update will have the silkscreen outline at +-1.27mm as proposed by @fruchti since I figured that having the silkscreen under the part body isn't that bad of big parts such as these style of pin headers?

Any comments on the triangle indicating pin 1?

atoav:
Alright, forgot about these : /
I think having the silkscreen covered is acceptable for pinheaders.

The triangle could also be a dot, but I don't mind – either thing is good.  

For the IPC thing I can't help you, I think if it is reasonable and people with experience think it is alright, it might be "okay enough" for now.

fruchti:
I really like the triangle. It should extend well to other types of packages where it provides a precise indication.

Another idea for the outline: We could have one bevelled corner at pin 1:
```
+----------+
| male     |
\----------+
/----------+
| female   |
+----------+
```
This would be a subtle difference between male and femaleheaders: When plugged together, the bevelled corners should match up. It also kind of shows the mirrored pin counting in the case of multi-row pin headers. On the other hand, there would be two pin 1 indicators with this, which is not that great.

atoav:
I like the idea of @fruchti too. I think pin 1 indicators on pinheaders are not as important as on a IC package, where it affects assembly. Pin indicators for pinheaders are typically important to find out which pin on the schematic corresponds to which pin on the PCB when troubleshooting.

fruchti:
Pin 1 indicators are extremely important for servicing and should be very clearly visible. IMO, one bevelled corner isn't prominent enough for that purpose, because after assembly it looks more like 1 missing corner, which could also be caused by a via in that location.

atoav:
Must say that in all the PCBs I ever saw I never saw a single one with Pin 1 indicators for pinheaders. And whenever I had issues where its location was important that was typically something where I had to look at the traces and the schematic anyways.  

That beeing said, I really don't care if it is there or not, so let's just add it : )

carrotIndustries:
> Must say that in all the PCBs I ever saw I never saw a single one with Pin 1 indicators for pinheaders.

I've definitely seen them one or multiple times on what must have been consumer electronics / power supply PCAs. That's where I got the idea from.

> Another idea for the outline: We could have one bevelled corner at pin 1:

While this may work, it's quite arbitrary and not recognizable without prior knowledge of this convention.

In case both sockets and pin headers end up having the same silkscreen: How about using the same package for both sockets and pins? It's still possible to have the appropriate 3D model by specifying the appropriate one in the part.

atoav:
Your arguments regarding the Pin 1 markers are convincing. Let's roll with it.  

I wouldn't make these the same package, just for consistency reasons (and you can be sure the right 3D-Model is linked in for the right part).

fruchti:
Yes, when choosing a female header part, accidentally ending up with a male 3D model shouldn't be possible.

atoav:
I created a way of generating male pinheaders with arbitrary spacing and pin sizes via SolidPython, OpenSCAD and FreeCAD. This works as follows:

1. Define the geometry and colors in SolidPython via translation, union, interesections all within a for-loop and export it to individual `.scad`-files
2. Convert the SCAD-file to STEP with FreeCAD

I still have a issue with the colors getting lost in FreeCAD and for now I need to manually fix this. I will try to get around this tho.  

One question came up: where is the pivot for pin headers? In the center of the bounding rectangle? Or the center of pin 1?

atoav:
Oh, I didn't think about kicad – right, lets use what can be used : )

carrotIndustries:
bf590b2c50ebe5943d1d1c59c0a935a172134fe7 adds the program used to generate these. instead of generating random UUIDs, it uses fixed UUIDs so the packages can easily be regenerated.

carrotIndustries:
We have these since 187c3040eabcdf270a3ea26c01ea9b3d2037ee47

atoav:
Good catch, I forgot to close this one

PR #30, opened 2018-04-06
Added Opamp Entities, Symbols and Units

atoav:
I am in the process of creating many opamp parts for the pool. It makes sense to implement units and symbols for them. The Enteties for Dual and Quad are pretty straightforward. Single is kinda generic, because some single opamps have additional offset inputs and I left it out here. I am not sure if the Dual Power Unit (for bipolar power) brings any real value here, but because the single power unit was called just that I thought I might add it too.


carrotIndustries:
The "Single Power" unit is supposed to be complemented by a "dual power" unit that has V+/V-/GND.

If you're fine with these changes, I'll merge this PR.

atoav:
Completely okay with it, go ahead!

PR #31, opened 2018-04-08
Added PJ301M-12 (3.5mm mono jack)

atoav:
Details were taken from [thonk](https://www.thonk.co.uk/wp-content/uploads/2014/02/Thonkiconn_Jack_Datasheet.pdf) and from the [manufacturers website](http://qingpu-electronics.com/en/products/WQP-PJ301M-12-306.html).


carrotIndustries:
- The symbol is missing the pins
 - The gate is supposed to be called 'Main'
 - Tags should be all lower case
 - each package should go in its own directory (to accommodate optional padstacks)
 - there is no hole in the center (use the 'circular mechanical hole' padstack)
 - the datasheet specifies rectangular holes, horizon supports oblong holes, so use these

atoav:
The thing with the hole... I used this part many times and never on a PCB with a hole (sometimes even with traces underneath, although on my own designs I avoid doing that). In dense designs you would actually use the opposing side for SMD parts and in that case a hole would be bad, that is why I left it out. But I will add it.

The pins are rectangular (1.0mm x 0.2mm) I think round holes are easier to manufacture and they give you the plus that you have a little bit of "wiggle room" which is very welcome if you have to fit these jacks to a panel or a case with a tight fit (and you typically should do this _before_ soldering). Quingpu for a weird reason has three different oblong hole sizes (0.5x1.5, 0.6x1.5 and 0.6x1.3).

atoav:
Okay I thought a part in the official library should fit the datasheet. I also modelled it in 3D which is always useful for big parts.

carrotIndustries:
- pins should be named direction-insensitive: signal or tip may be good candidates
 - connector should be facing to the right
 - place text top/bottom of symbol (see new coaxial connector)
 - silkscreen near pin 3 needs some work (lines cross)
 - package outline should be polygon
 - name the center pad M1
 - courtyard spacing is inconsistent
 - superfluous pad parameters (pad diameter)
 - assy refdes should be placed at the short edge of the package, feel free to make it bigger (J1234) still has to fit

PR #32, opened 2018-04-08
Added P0915N Package for 9mm potentiometers

atoav:
This is a commonly used package for potentiometers and encoders (used by TT Electronics, Alps, etc). As a reference [this datasheet](http://www.ttelectronics.com/sites/default/files/download-files/Datasheet_PanelPot_P09xSeries.pdf) and real parts for sanity checking were used


carrotIndustries:
- Swap group can be 0 everywhere: 0 indicates 'not swappable'
- only one pin name is visible. is this intentional?
- the symbol has not-so-well fitting text placements for some orientations
- I'd prefer RV for prefix
- the gate should be named 'Main'
- the symbol/unit/entity should be called 'Potentiometer', the other generic units etc. also don't have 'generic' in their name
- create one base part without value and use 'create part from part' to create parts with value
- always link to the manufacturer's datasheet, not the mirror of some distributor
- the mounting pads are unconnected. the use may wish to connect these to GND, so we need a need a unit called "Potentiometer with shield".
- package.json is misplaced (see other PR)

atoav:
3D model was again modelled by me.

carrotIndustries:
> 3D model was again modelled by me.

As far as I can tell it's definitely derived from the manufacturer's model, so better not include it in this repo. Maybe it's best to create a separate repo for 3D models of dubious license.

atoav:
Just checked (also for these switches) and you're right. Both were part of my private collection for a few years and for some reason they ended up in the wrong folder. This proofs that your proposal for 3D models for parts with dubious licenses makes sense.    

I might end up doing these models on my own from scratch at some point in the future, because I use them practically in every other project. I did jobs as a 3D-modeller all the time since, so this is not really the issue.

atoav:
Took me some time to model this, because I had some issues with the kicad importer (the meshes have to be solid with no holes and manifolds, with quite tight tolerances, which was a bit of a challange to pin down).

niclashoyer:
link to datasheet changed: https://www.ttelectronics.com/TTElectronics/media/ProductFiles/Potentiometers/Datasheets/P09x.pdf

carrotIndustries:
Sorry for letting this sit for so long, but the text placements need to be updated to support multi-line values.

2019-05-19-165152_411x281_scrot

atoav:
I have a question about this. I don't really know where the point of reference for that multiline value _is._ 

Should I adjust it so all lines fit neatly? Should I adjust it's center ("Group")? Or its top ("Value")?

This quesion essentially boils down to the Question how it will look if it is a single line:
```
$VALUE         $VALUE
Group    =>
Tag
```
or
```
$VALUE     
Group    =>    $VALUE
Tag
```

A bit confusing for me, sorry.

carrotIndustries:
> This quesion essentially boils down to the Question how it will look if it is a single line:

It'll look as in your first case.

To try this out, change `$VALUE` to something else so that it doesn't get replaced in the preview window.

atoav:
Okay hope this is alright now

carrotIndustries:
> Okay hope this is alright now

Perhaps I should have elaborated on the value placement a bit more:

The value should be placed such that it can grow to any number of lines without obscuring other parts of the symbol. It's also preferred to have the `$REFDES` text above the symbol (positive Y)

atoav:
Aha, that means it should actually be placed outside of the bounding box of the symbol. The reason why I placed it under the symbol and not above is, because any growing text would very likely collide with the nets that you would very likely draw. Any suggestion how to solve this?

2019-05-19 20_25_28-Symbol preview

PR #33, opened 2018-04-09
Added Bridge Rectifier Symbol, HD DIP package and CBRHDSH2-40

atoav:
The symbol design oriented itself from the original diode schematic, but may differ a little bit in size and proportions of the diode symbol.
The HD DIP package seems to be a special package used by Central Semi, so I was not quite sure how to place it in the folderstructure.
The CBRHDSH2-40 and its package have been drawn in accordance to [this datasheet](https://www.mouser.de/datasheet/2/68/cbrhdsh2-40-17840.pdf)


carrotIndustries:
- CBRHDSH1-* should inherit from a common base part
- tags should be all lower case
- pins should have nonzero length, see image below 
- the long line indicating pin 1 should be straight


2018-04-25-015737_981x502_scrot

PR #35, opened 2018-04-12
Added Trimmer Unit, Symbol, Entity and Packages

atoav:
Trimmers with package 68W are TT Electronics 12-Turn Precision trimmers (footprint derived from Datasheet) and the Packages 329XX are Bourns (with same pads/holes, but different placement of the adjustment screws)


atoav:
@carrotIndustries I made a 3D-model for the 68W package, but whenever I try to load it the editor crashes. Freecad finds no error with the mesh and I tried multiple things, but due to the shadowy nature of these crashes I have a hard time pinning things down. So this may be a Bug or I am doing something wrong – would you look into it if you have time?

atoav:
Moved PR to https://github.com/carrotIndustries/horizon-pool/pull/36

PR #36, opened 2018-04-16
Added Trimmer Unit, Symbol, Entity and Packages

atoav:
Added Trimmer Unit, Symbol, Entity and Packages (Referencing https://github.com/carrotIndustries/horizon-pool/pull/35)


carrotIndustries:
- refdes should always be above value
 - 3D model license? see other PR
 - in some packages, the silkscreen outline is obscured by the part body

atoav:
I will fix the rest next week, when I am back in town.
I made the 3D-models from scratch so licensing is no issue.

PR #38, opened 2018-04-25
Added TL072 opamps (TSSOP, SOIC, DIP)

atoav:
Parts based on the datasheets included


carrotIndustries:
- some parts duplicate the tags they're inheriting
 - always link to the manufacuturer's website for datasheets
 - use base parts to so you'll only have to specify the datasheet link once

atoav:
One question, how many levels of nesting are acceptable?  

For now I have a base part for each package and all the parts that have that package inherit from it (these are parts with different datasheets and manufacturers). Should I create a base part for every unique datasheet?

carrotIndustries:
> One question, how many levels of nesting are acceptable?

Technically, there's no limit. So do whatever fits your particular use case

In this case, the hierarchy should be this way (feel free to object)
 - generic dual so-8 opamp (many share the same pinout)
   - TL072 by TI
     - TL072ABC
     - TL072CYZ

> Should I create a base part for every unique datasheet?

exactly

atoav:
Okay, that makes sense – I just hesitated to overly nest things 👍

carrotIndustries:
these have CRLF line endings, please use unix-style line endings

carrotIndustries:
- units/entities still have CRLF line endings
 - some part descriptions contain line breaks

atoav:
I double checked, there should be no CRLF line endings anymore

carrotIndustries:
The DIP base parts reference the socketed package. Is this intentional?

atoav:
> The DIP base parts reference the socketed package. Is this intentional?

Nope. Good catch

carrotIndustries:
> Edit: should be resolved

The DIP-8 part still references the socketed one.

Perhaps the socketed packages should be an alternate package for the non-socketed ones, in order for the socketed packages to be used at all.

atoav:
There are:
1. DIP8_7.62mm_lead_span_2.54mm_pitch
2. DIP8_7.62mm_lead_span_2.54mm_pitch_(diy_silkscreen)
3. DIP8_7.62mm_lead_span_2.54mm_pitch_(with_socket)
4. DIP8_7.62mm_lead_span_2.54mm_pitch_(with_socket_diy_silkscreen)

Package 1 is obviously going to be the main package. I am not sure whether I really understand how alternate packages work. I know I can select one alternate package per package, but wouldn't there be 3 alternate packages per package? I could of course buld some sort of circle `1 → 2 → 3 → 4 → 1` but I am not sure this is how it was supposed to work.

Just wanted to ask before I change all dip packages and then find out I did it the wrong way

atoav:
Bot!

atoav:
Bot!

atoav:
I did another pass of checking this one again and to my eyes it is okay. 
The *DIY Silkscreen* package variant fails, because the indicating silkscreen collides with the package, but in my eyes this is acceptable in this case:

image  

@frmdstryr @carrotIndustries I think merging that commit would be good, because of the packages alone. DIP is kinda oldschool, I know, but for a certain crowd (e.g. in education) it has not lost it's importance.

RX14:
What's the human readable description of the diff to the packages? It's really hard to review package file diffs.

I think it's probably not a good idea to put more than "TL072CD" in the MPN. Both the manufacturer and the package's full name are visible next to the MPN right in the part browser and pool manager, so having multiple parts with the same MPN isn't a bad thing. It's very easy to find the right MPN in the list and then scan to the package and manufacturer columns for the exact part.

PR #41, opened 2018-05-02
Added SPDT switches and subminiature parts

atoav:
Two units/entities/symbols for SPDT switches (ON-ON and ON-OFF-ON) as well as two subminature switches by TT Connectivity see [here](http://www.te.com/usa-en/product-2267078-2.html) and [here](http://www.te.com/usa-en/product-2267078-5.html)


carrotIndustries:
- pin direction should be passive
 - inconsistent pin name visibility
 - 3d model appears to be derived from manufacturer's model of unknown license
 - inconsistent paths: "/passive/mechanical" should be "/mechanical"

carrotIndustries:
- Silkscreen refdes size should be 1mm
- DPDT package uses SPDT 3D Model
- package names should be less generic (use MPN)
- symbol text placement needs to support multi-line text
 - DPDT symbol is a bit confusing, why is the vertical line interrupted?

atoav:
> DPDT symbol is a bit confusing, why is the vertical line interrupted?

This was meant to be a "dashed" line, maybe I should make the dashes smaller?

carrotIndustries:
There were still some issues with the symbols, see my commits for the changes implemented.

The text placement for the SPDT switches still needs some wotk.

atoav:
Okay this looks good now, the only thing missing is a 3D-model for the dpdt variant, but I might at that one later in a seperate PR

PR #43, opened 2018-05-03
Added Vactrol optoisolator and LCR-0202

atoav:
Added symbol, unit and entity for vactrol (a optoisolator built from a LED and a LDR). Also added the [LCR-0202 by Nanyang Senba Opto Elec](http://en.nysenba.com/upfiles/file/13521047800.pdf). Their server seems to be on fire, so [here is an alternative datasheet](http://p.globalsources.com/IMAGES/PDT/SPEC/684/K1051105684.pdf) from a retailer.


carrotIndustries:
- symbol should be left to right
 - pins are too short, use default length
 - all pins should be passive
 - use A/K for diode pins
 - don't use dot for pin 1 indication
 - 3d model path is wrong
 - silkscreen is pretty but to elaborate to my taste

atoav:
I used the dot as a pin 1 indicator, because that is what the package itself uses. What else do you suggest?

carrotIndustries:
Symbol texts should be below/above symbol body: 
2019-05-19-173724_246x244_scrot

Orientation-specific test placements aren't really necessary for this kind of part.

Perhaps rename Vactrol to something more universally-understandable such as resistive opto-isolator.

carrotIndustries:
36fecf0:
 - remove alt names, fix swap groups
 - fix text placement

atoav:
Huh, thanks for the extra legwork, had to go to sleep earlier yesterday!

PR #44, opened 2018-05-03
Added Dual NPN/PNP, SOT-363 and BC85x6AS

atoav:
Added SOT-363 package, Dual NPN and Dual PNP entities as well as two parts which make use of both ([BC856AS](https://www.diodes.com/assets/Datasheets/BC856AS.pdf) and [BC846AS](https://www.diodes.com/assets/Datasheets/ds30833.pdf))


carrotIndustries:
- silkscreen text is to close silkscreen graphics
 - pad position doesn't match datasheet
 - for generic packages, use the kicad packages for reference since these are real-world proven instead of some random manufacturer's recommendation or create the package specific for this manufacturer

atoav:
Huh, not sure what went wrong here. I must have been in a confused state of mind. These parts don't use the SOT-323 package as I wrongfully wrote, but the SOT-363 package. On top of that this thing was filled with mistakes. Good catch, thanks.

I am not sure how to check the kicad packages as I am not really a avid kicad user, but based on the added 3D-model the footprint now makes sense.

atoav:
Okay this should work now

PR #46, opened 2018-05-08
Added Comparator Units, Symbol, Entities and multiple parts

atoav:


carrotIndustries:
- this particular family of comparators has open-collector outputs, this isn't reflected in unit and symbol
 - don't tag base parts 'base' so non-base parts can inherit their tags

carrotIndustries:
- Gate names are messed up (see opamp)
 - input pins on symbol should be consistent with opamp
 - remove generic tag on parts

atoav:
This should work now

carrotIndustries:
- Base parts reference socketed dip package, should reference non-socketed package
 - DIP package fixes should go in a separate PR

PR #48, opened 2018-05-21
Added many yageo resistors

atoav:
I created parts for yageo's smd resistor series. For 0805 the RC (General Purpose), RT (High Precision, High Stability) and RE (Thick Film Precision Grade) resistors, as well as 0603 RC (General Purpose) series.  

The parts have been created automatically via python by parsing the tables [on the manufacturers site](http://www.yageo.com/portal/product/product.jsp). Any suggestions?


carrotIndustries:
- MPN should be just the MPN
 - see the CPL parts on how the value is to be formatted
 - include parametric data (see CPL parts)

atoav:
- Removed info from MPN
- Formated values according to CPL
- Added Parametric Data

**Please note:** all parts got new UUIDs, I was lazy and just generated a new patch.

My rationale for using `100R` instead of `100 Ω` were two ideas:
1. a smeared omega can be more easily mistaken for a zero, than a R. For the same reason values like 4.7k are sometimes written like 4k7 (a dot can be easily overlooked or not printed)
2. when searching for a resistor it is easier to type "100R" than "100 Ω"

**Point 1** is definitly better with the space added between unit and value. In the end this is a matter of clear typography. We could consider using a [slashed zero](https://en.wikipedia.org/wiki/Slashed_zero) throughout the software, as well as a bolder floating point to avoid these mistakes.  

**Point 2** can be fixed by using the parametric data I suppose..

atoav:
I added parametric data and changed the values to the ohm notation. These are many resistors, but because I auto generated them for the most part, they should not differ except for values and names etc.

atoav:
I also added the TH resistors from #49

carrotIndustries:
With the concrete parts thing right around the corner, we can definitely consider merging this.

Some thing caught my eye though:
 - Some descriptions have trailing spaces
 - Parts don't have the base 0603/etc parts set as base parts. That'll prevent setting the appropriate `concrete_for` part.

atoav:
Should we add a base part for 9mm TH resistors?

fruchti:
Some comments regarding the TH resistors:

* The TH resistor packages are commonly referred to as 0204, 0207, 0309 (for 1/8, 1/4 and 1/2 watt resistors). These codes should be added to the package names for searchability.
* For horizontally mounted TH resistors, the lead spacing should be visible in the part name. It might also be nice to offer different lead spacing variants, but this is probably too much for now.
* The silkscreen text of the TH resistor with its 1.3 mm height is currently larger than the convention's recommendation of 1 mm. With a smaller text, the values should also fit within the resistors.
* Why not make vertical and horizontal alternative packages? Currently there are two parts for each resistor without different MPN.

atoav:
> The TH resistor packages are commonly referred to as 0204, 0207, 0309 (for 1/8, 1/4 and 1/2 watt resistors). These codes should be added to the package names for searchability.

Do you have any resource on that so I get this right?

> For horizontally mounted TH resistors, the lead spacing should be visible in the part name. It might also be nice to offer different lead spacing variants, but this is probably too much for now.

I had this as part of the package – MPN: `Generic Resistor 6.3 mm Length`. I could just add another lead space variant as an alternate package?

> The silkscreen text of the TH resistor with its 1.3 mm height is currently larger than the convention's recommendation of 1 mm. With a smaller text, the values should also fit within the resistors.

Indeed, good catch, my next commit will fix this

> Why not make vertical and horizontal alternative packages? Currently there are two parts for each resistor without different MPN.

You are right – I don't remember why I didn't do that, but it seems to work just fine.

fruchti:
> Do you have any resource on that so I get this right?

This seems to be part of a DIN standard, although I couldn't find it directly so far. For other resistor series, this size code is part of the MPN (e.g. https://www.vishay.com/docs/28766/mbxsma.pdf).

> I had this as part of the package – MPN: Generic Resistor 6.3 mm Length . I could just add another lead space variant as an alternate package?

I thought that referred only to the resistor body length and haven't actually checked. “Lead spacing” or something alike might still be clearer for the pad-to-pad distance (between cemtres).

> Indeed, good catch, my next commit will fix this

The texts could also still be moved downwards a bit while still having enough clearance between text and silkscreen outline. Might help if there are edge cases where the value doesn't fit.

atoav:
> I thought that referred only to the resistor body length and haven't actually checked. “Lead spacing” or something alike might still be clearer for the pad-to-pad distance (between cemtres).

Forget that, I lied. This was indeed the body length of the resistor (because of the 3D-model). I added it to the package name. Any other lead spacings I should add as alternate packages?

This DIN thing still confuses me, because the vishay datasheet you linked actually has 0.4 watts and 0.6 watts for the 0204 and 0207, does this mean the body size is actually the thing I should use as orientation?

DIN Code | Power rating | body length | body diameter | lead diameter
-- | --- | --- | --- | ---
0204|1/8 (0.125) | 3.0 ± 0.3 | 1.8 ± 0.3 | 0.45 ± 0.05
?|1/4 (0.25) | 6.5 ± 0.5 | 2.5 ± 0.3 | 0.6 ± 0.05
?|1/2 (0.5) | 8.5 ± 0.5 | 3.2 ± 0.3 | 0.6 ± 0.05
0414|1 | 11 ± 1 | 5 ± 0.5 | 0.8 ± 0.05

fruchti:
> This DIN thing still confuses me, because the vishay datasheet you linked actually has 0.4 watts and 0.6 watts for the 0204 and 0207, does this mean the body size is actually the thing I should use as orientation?

Sorry, I should have been more clear. The standard is probably from a time where resistor size and power rating were a strict 1:1 relationship. Nowadays, the two dimensions in the code seem to be rounded millimeter dimension. How exactly the rounding works or if the resistors are just nowadays made slightly smaller, I don't know (since I don't have the standard at hand).

> Any other lead spacings I should add as alternate packages?

On PCBs with a lot of DIP stuff, an even multiple of 0.1" looks nice, IMHO. I like to use 400 mil = 10.16 mm, since 300 mil are often quite tight for 0207 resistors.

atoav:
Okay added the 400 mil variant incl. 3D model too.

For the DIN code, the most convincing theory I found was that the code describes the next-smaller variant that fits into the space (so basically the round up integer value of the measurements of the resistor body).

I added 0307 to my resistor packages and moved the package directories accordingly

fruchti:
> Okay added the 400 mil variant incl. 3D model too.

I completely forgot that the 3D model would have to be adjusted as well. Thanks for all the work!

> For the DIN code, the most convincing theory I found was that the code describes the next-smaller variant that fits into the space (so basically the round up integer value of the measurements of the resistor body).

I looked around a bit more but couldn't find anything concretely. It's likely that IEC 60115-2 contains the relevant information, but I lack access to it. The KiCad repositories don't show any reasoning, but have the naming I'm used to: https://github.com/KiCad/kicad-footprints/tree/master/Resistor_THT.pretty, so I'd keep it to 0207 even if the rounding was inconsistent.

Interestingly, the lead spacing is always a multiple of 2.54 mm in the KiCad repositories. Out of curiosity: Where did your 8.75 mm come from?

atoav:
> Where did your 8.75 mm come from?  

Interesting question. It was quite a while ago, so I don't remember, but it might have been the datasheet. But now that you mention it, it should definitly fit the raster.

I did a quick check with some parts, calipers and a rastered pcb:
- 7.62 mm (3x 2.54 mm or 4 holes long) works fine if you bend the legs sharpely
- 10.16 (4x 2.54 mm or 5 holes long) works fine as well

I will change all the 8.75 mm ones to 7.62 mm to fit the raster better.

niclashoyer:
soo, is this a thing now?

Bot!

carrotIndustries:
> soo, is this a thing now?

the bot only listens to prefixes, i.e. your post has to start with Bot!

niclashoyer:
Bot!

niclashoyer:
hmm, I suspected that, given that there are many many parts in here:

Bildschirmfoto von 2021-01-05 01-12-14

still a great job on the actual review process! btw: Is the source for the docker container somewhere?

carrotIndustries:
> still a great job on the actual review process! btw: Is the source for the docker container somewhere?

Of course: The bot itself is ~1000 lines of spaghetti code: https://github.com/horizon-eda/horizon/blob/master/src/pr-review/pr-review.cpp with some rather lengthy SQL queries. It gets built by this dockerfile: https://github.com/horizon-eda/horizon/blob/master/Dockerfile-pr-review and is used by this workflow: https://github.com/horizon-eda/horizon-pool/blob/master/.github/workflows/review.yml

Also check out that blogpost: https://blog.horizon-eda.org/progress/2020/10/29/progress-2020-09-10.html

carrotIndustries:
Now that passives are in their own pool, these resistors would be better off in their own pool as well.

RX14:
I really hope this makes it into an official pool - I have many yageo resistors manually added to my designs.

RX14:
I see the script used to generate these isn't included in this PR? If I had a copy I could work on this. Though I'd probably drop the TH parts.

atoav:
Yes, @RX14 I made these before there policy of including the generative scripts existed. I still have it lying around somewhere though. I will deal with this on the weekend (I hope). Where for many people the pandemic meant more time, for me it is the other way around.

RX14:
Thanks! Though the offer's still there to finish it off myself.

atoav:
I had a very dense/filled time the last year, so I could not get around to this.

I moved the through-hole parts into their own pool/repo (see: https://github.com/atoav/horizon-pool-yageo-th-resistors). IS there a mechanism to include them in the "avialable" section of horizons pool settings?

If anybody finds the time to help with the other parts in this PR, feel free to do so

frmdstryr:
> IS there a mechanism to include them in the "avialable" section of horizons pool settings?

Add it to the index https://github.com/horizon-eda/horizon-pool-index

carrotIndustries:
No that's it's in the index, can we close this PR? One thing that might be worth it is moving the packages and base parts to the main pool so other pools can build on them without having to depend on the this specific pool.

atoav:
> One thing that might be worth it is moving the packages and base parts to the main pool so other pools can build on them without having to depend on the this specific pool.

I think this is a good idea. I look whether I can get around to creating a new PR for that today.

PR #50, opened 2018-06-18
Added QSOP-16, dual power and THAT2162 IC

atoav:
The QSOP-16 Footprint is based on the one specified [here](https://www.diodes.com/assets/Package-Files/QSOP-16.pdf). The 3D-model comes from kicad and again I had to re-export it with Freecad to avoid crashes.  

The THAT2162 (a integrated Voltage Controlled Amplifier) was made in accordance to [this datasheet](http://www.thatcorp.com/datashts/THAT_2162_Datasheet.pdf). Because it was the first bipolar IC that needs Vee, Vcc and GND I added the Dual Power Unit and Symbol. Feel free to suggest a different design or approach.


carrotIndustries:
Looks like you forgot the unit 70bbe...  in this PR

atoav:
Okay I fixed it now

carrotIndustries:
Any reason for placing both refdes and value below the main symbol?

The dual power symbol should look a bit different:

- Use rectangular outline
- Add refdes/value text

atoav:
Bot!

PR #51, opened 2018-06-26
Add ULN2803ADW

fruchti:
It's been a while since I've been able to contribute and I've probably forgotten half of the dos and don'ts, so feedback is very welcome ;)


fruchti:
Thanks for the feedback! I hope i resolved everything. Some questions/comments:

> Keep silkscreen outline as simple as possible - delete 'hooks' on the bottom one

Is there a hard limit here? Because in the IPC documents linked in past issues/PRs, these are included once they are larger ([here](http://www.pcblibraries.com/downloads/temp/PCB_Design_Optimization_Starts_in_the_CAD_Library_7126027.pdf), page 139 or [here](http://www.ocipcdc.org/archive/What_is_New_in_IPC-7351C_03_11_2015.pdf), page 20), but I didn't find an absolute length “decision boundary”. Only clearances for the lines are given.

carrotIndustries:
> Is there a hard limit here?

Not really. By not including the 'hooks' there's one less variable to worry about.

fruchti:
> Not really. By not including the 'hooks' there's one less variable to worry about.

So you would include them never for SOP-style packages? There are some where the body length exceeds the distance between the outermost pads by quite a bit, so including them wouldn't be a bad idea. [Here's](http://www.onsemi.com/pub/Collateral/492B-01.PDF) an example, but there are probably better ones.

atoav:
> It's been a while since I've been able to contribute and I've probably forgotten half of the dos and don'ts

I know that feeling. This is why we need to:
- move needed information as close as possible to the parts in the software that need it
- add rules for things that can be easily checked (e.g. whether $REFDES is above $VALUE in a symbol)
- add sensible defaults so that certain things don't have to be thought about at all anymore
- add "macro commands" that again help you to avoid missing things (e.g. a command that adds all needed text labels like $REFDES and $VALUE at once in the right order)

I will detail how I think this could be done in a issue at the horizon repo soon

carrotIndustries:
Bot!

carrotIndustries:
Bot!

carrotIndustries:
>    x Tag "driver," doesn't match regex
>     warning Only gate must be named "Main"

good bot

fruchti:
I fixed the issues, added a 3D model and rebased on top of the current master.

> > x Tag "driver," doesn't match regex
> > warning Only gate must be named "Main"
> 
> good bot

Good Bot! This is great feedback! One thing it didn't catch was the wrong assembly layer text size (was 1.5 mm, should be 1 mm).



> > SO18, 7.62 mm lead span
> 
> doesn't match measured pitch of 9.400 mm

I don't remember what definition of lead span I used here, but the pad pitch makes a lot more sense, so I changed it.

fruchti:
The assembly text in the preview doesn't seem to have been resized and moved?
2020-09-21-120746_1279x803_scrot

carrotIndustries:
> The assembly text in the preview doesn't seem to have been resized and moved?

Looks as expected to me. This is probably caused by your browsers cache. Since other users and I were seeing similar behaviour I fixed the bot to make sure this doesn't happen again.

fruchti:
> > The assembly text in the preview doesn't seem to have been resized and moved?
> 
> Looks as expected to me. This is probably caused by your browsers cache. Since other users and I were seeing similar behaviour I fixed the bot to make sure this doesn't happen again.

Indeed! I thought since I renamed the package file, the cache wouldn't be an issue, but of course the image file name is based on the UUID, which didn't change. Thanks for the fix!

PR #52, opened 2018-07-19
Padstack for through-hole with rectangular pad

endofexclusive:
Circular hole with rectangular pad. Can be used to highlight pins on through-hole components.


fruchti:
I've opened [an issue](https://github.com/fruchti/horizon-library-convention/issues/4) for discussing if we want rectangular pin 1 pads in general.

carrotIndustries:
squash merged in 59c2a4459a003c0219246db8c31baf753f5189f6

PR #54, opened 2018-07-30
Add ON Semi NVTFS5C478NL

fruchti:
I divided the solder paste polygon for the large pad into 4 smaller polygons so it takes the shape of [ON Semi's recommendation](http://www.onsemi.com/pub/Collateral/AND9138-D.PDF) for a paste contraction > 0.

I didn't find a freely available 3D model for the package, so it doesn't include one.


carrotIndustries:
Looks good overall, but replace lines on assy/package layer with polygons

carrotIndustries:
Adjusted assy refdes size so Q1234 would fit

PR #55, opened 2018-08-01
Add NSR0170HT1G schottky diode

fruchti:
I created a schottky unit, symbol and entity for this. I believe this is the correct way, since the schottky symbol shouldn't be an alternative symbol for a normal diode.

The 1.2 mm dent in the assembly layer for the SOD-323 package doesn't look to great, though. Should we stick with this value even for very small packages?


carrotIndustries:
> The 1.2 mm dent in the assembly layer for the SOD-323 package doesn't look to great, though. Should we stick with this value even for very small packages?

That's true, half the length of the shorter side of the package seems like reasonable maximum size for the dent. Your thoughts?

fruchti:
Sounds good. I've updated the package and also reoriented the text in the assembly layer similar to the other packages.
I'll update that in the convention, too. I suggest that the text should be downscaled in 0.1 mm steps until a reference designator with 4 numerical digits fits.

PR #56, opened 2018-08-11
Add ECS-MPIL0530 SMD power inductors

fruchti:
Two questions in this context:
- Is there any way to make “EU inductor with core” the default symbol? I'm not sure why they are alternatives anyway.
- Should non-polarised components include the pin 1 indication in the assembly layer? Manufacturing-wise it doesn't make much sense, but maybe it's helpful for probing around prototypes or something?

### DIFF
--- a/packages/passive/magnetics/ECS-MPIL0530/package.json
+++ b/packages/passive/magnetics/ECS-MPIL0530/package.json
@@ -169,12 +169,6 @@
                 -1525000
             ]
         },
-        "252a435f-3bab-4489-a5c0-ce5f036bca01": {
-            "position": [
-                2745000,
-                2590000
-            ]
-        },
         "307b72f8-da59-47ea-9c87-491786294b69": {
             "position": [
                 -3020000,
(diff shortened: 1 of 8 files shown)


carrotIndustries:
> Is there any way to make “EU inductor with core” the default symbol? I'm not sure why they are alternatives anyway.

Where would you expect that option? Part or user setting?

> Should non-polarised components include the pin 1 indication in the assembly layer? Manufacturing-wise it doesn't make much sense, but maybe it's helpful for probing around prototypes or something?

Since the symbol has no visible pin names, that doesn't seem all that helpful.

fruchti:
> > Is there any way to make “EU inductor with core” the default symbol? I'm not sure why they are alternatives anyway.
> 
> Where would you expect that option? Part or user setting?

I think it'd make more sense if these were separate and not alternatives, actually. Alternatives, when they exist should be all equally sensible in all cases, so a preference setting probably isn't that useful after all.

carrotIndustries:
So your sugggestion is to add a separate unit/entity for inductors with and without core?

PR #57, opened 2018-08-28
Add STM32F405RGT

m3mory:
Added STM32F405RGT, TQFP-64 footprint is generated to the specifications in the STM32F405 datasheet


fruchti:

PR #59, opened 2018-10-09
Add WS2812B

fruchti:
Based on this [datasheet](http://www.world-semi.com/DownLoadFile/108). I looked around for a 3D model for quite a bit, bit didn't find any with a compatible licence, so there is none included. I took the liberty to change some things in comparison to the datasheet:

* The pin numbering is rotated by 180 degrees. IMHO, it makes more sense that pin 1 is actually the pin with the bevelled corner of the package. This seems to be the convention for all the other LED PLCC footprints I found, too.
* DIN/DOUT have been renamed to DI/DO, so they take less space in the symbol.

### DIFF
--- a/entities/opto/WS2812B.json
+++ b/entities/opto/WS2812B.json
@@ -1,0 +1,19 @@
+{
+    "gates": {
+        "79a5fea5-ea4b-47ac-a841-2b62d92ea338": {
+            "name": "",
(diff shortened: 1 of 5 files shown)


atoav:
Merry XMas @fruchti, decided to  build a (crude) 3D-model. See #74

PR #63, opened 2018-10-30
Added Two-Terminal Bicolor LED Entity and Lumex SSL-LX30FT14IGW

atoav:
The Datasheet: [http://www.lumex.com.tw/spec/SSL-LX30FT14IGW.pdf](http://www.lumex.com.tw/spec/SSL-LX30FT14IGW.pdf).  

The 3D-model is by me, I made it for another part, but it fits the Datasheet.


carrotIndustries:
- Create a new symbol/unit/entity that contains two antiparallel LEDs to indicate what's there. Pads should have short names such as 1/2
 - MPN should be just the MPN
 - Package looks pretty similar to other 3mm LEDs, why a new package?

atoav:
Okay I replaced that specific package with a more generic 3mm round top version and added multiple 3d-models (all based on [this one by kicad](https://github.com/KiCad/kicad-library/blob/master/modules/packages3d/LEDs.3dshapes/LED_D3.0mm.FCStd)) for multiple colors. This should work as a base for most other 3mm LEDs out there

In addition to that I created a antiparallel bicolor led unit, symbol and entity and cleaned up the parts MPN

PR #64, opened 2018-11-02
Added unit, symbol and entity for SPSPT push switch and D6R series by ckswitches

atoav:
The datasheet is here: https://www.ckswitches.com/media/1341/d6.pdf  

For the 3D portion I relied on both the datasheet and measurements from a real part. I added multiple color variants, each with the according MPN


atoav:
Changed as requested

carrotIndustries:
> $VALUE should be below the symbol
Still not there.

$VALUE is supposed to go below the symbol since it may get replaced by multi line text in the future for further information.

atoav:
AH below the _symbol_! For some reason my head parsed this as below the _$REFDES_. Makes more sense now, sorry for the hassle.

PR #65, opened 2018-11-02
Added Kingbright WP242 series (cylindrical 3mm LED)

atoav:
The series is here: http://www.kingbrightusa.com/KCpack.asp?txtPack=WP424&page=1

I added a "generic 3mm LED" package, because these are mostly the same. I added multiple 3D-models to it for the colors. Is this a good approach or should I make a seperate Kingbright-specific package?


carrotIndustries:
- MPN should contain just the MPN and nothing else
 - See package check results
 - 3d model for grey LED has wrong filename (only works on platforms that have case-insensitive file names)
 - generic package is fine as long as there are other manufacturers making LEDs of the same style (flat top package)
 - maybe add "flat top" to the package name, dome-style leds will then get their own package

atoav:
The only thing I didn't really "get" was the wrong filename, could you check and give me a pointer if something is wrong?

carrotIndustries:
- Filename seems to have fixed itself. 
 - Courtyard should include plastic part (fixed in https://github.com/carrotIndustries/horizon/commit/2eb2334cf722df5f01a5dbbcbddb675a94b4610a)

atoav:
Okay fixed that courtyard

PR #67, opened 2018-11-02
Added Alfa AS3320 IC

atoav:
This is a VCF (Voltage controlled filter) IC by the lithuanian manufacturer ALFA. 
Datasheet is here: http://www.alfarzpp.lv/eng/sc/AS3320.pdf

The AS3320 is a replacement for the out of production CEM3320 which is used in many analog synthesizers.

Feel free to deny this because it is a little oddball.

### DIFF
--- a/entities/ic/manufacturer/alfa/as3320.json
+++ b/entities/ic/manufacturer/alfa/as3320.json
@@ -1,0 +1,19 @@
+{
+    "gates": {
+        "2360662f-1f83-477e-a87c-e3d3bd1d1fde": {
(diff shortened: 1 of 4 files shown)


carrotIndustries:
See the latest commit for my suggestion on how the symbol should look like.
 - Power pins on top/bottom
 - consistent 2.5mm pin spacing
 - symbol rectangle is centered

I also removed all alt names since these are separated by spaces. See the Unit "PIC32MZ2064DAH176 B" for a usage example.

PR #68, opened 2018-11-03
Added SOIC-16 Package and Alfa AS3310

atoav:
This is again one of these Alfa ICs, this time a ADSR envelope generator. 
The datasheet is here: http://www.alfarzpp.lv/eng/sc/AS3310.pdf

This is again a replacement of the CEM3310 that went out of production.

I also added a SOIC-16 footprint, which was mostly copy, paste and move from the existing SOIC-14 footprint. I added a 3D model from KiCads library (see [this part](https://github.com/KiCad/kicad-packages3D/blob/master/Package_SO.3dshapes/SOIC-16_3.9x9.9mm_P1.27mm.step))


carrotIndustries:
- symbol (see other alfa ic)
 - entity should have prefix U
 - part should use non-socketed package, socketed package should be alt package for non-socketed package

atoav:
This package has two GND (one power ground, one signal ground). I still ended up packing the signal GND on the (left) side, for clarity. Is this alright?

PR #70, opened 2018-11-25
Added STM32F767ZI, STM32F446RET, STM32F042K6
 Updated STM32F405RGT

m3mory:
Added STM32F767ZI, STM32F446RET, STM32F042K6
Updated STM32F405RGT

### DIFF
--- a/entities/ic/mcu/stm/stm32f042k6.json
+++ b/entities/ic/mcu/stm/stm32f042k6.json
@@ -1,0 +1,19 @@
+{
+    "gates": {
+        "752964bf-c77e-407a-992a-838b3f41dc34": {
+            "name": "",
+            "suffix": "",
+            "swap_group": 0,
+            "unit": "407344b2-4f8e-452b-9c76-acc03381ac9b"
+        }
+    },
+    "manufacturer": "ST Microelectronics",
+    "name": "STM32F042K6",
+    "prefix": "U",
+    "tags": [
+        "ic",
+        "mcu"
+    ],
+    "type": "entity",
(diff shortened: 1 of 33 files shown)


fruchti:
Some feedback based on a first glance:

# Footprints

* Reference designators should be visible after assembly
* Package/assembly layers are missing
* The courtyard parameter program is missing
* The pin 1 indicator for the TQFP144 is unlike all the others
* Silkscreen clearances are incorrect
* Naming should not include the dash (“TQFP100” instead of “TQFP-100”)

You might look at the currently closest thing (that I know of) to a guide to the pool convention we have [here](https://github.com/fruchti/horizon-library-convention).

# Symbols, units, parts

* The units don't have all the alternative pin functions as alternative names. I like to have them because you can clarify the peripheral used for each pin immediately in the schematic (for example connecting a SPI bus to pins clearly labelled as SPI2_MOSI, etc.)
* For the symbols, there are some differences in the “padding” around the pin names (compare STM32F042 and STM32F405). Also, The STM32F042 symbol isn't symmetric and not properly centred
* Datasheets should link to the manufacturer's website, not Google
* The part description is necessary

FYI: I started (or tried to start) a discussion on where to place power pins in the symbols [here](https://github.com/fruchti/horizon-library-convention/issues/7) and on splitting parts into units (as you have done with the STM32F767ZI) [here](https://github.com/fruchti/horizon-library-convention/issues/8).

I was going to suggest something like this for the power pin placement (WIP): 2018-12-02-155412_2476x1496_scrot

fruchti:

PR #72, opened 2018-12-03
Added LME49720 Dual Opamps

atoav:
Not much inheritence going on here, there are only two parts with two MPNs. Datasheet is here: http://www.ti.com/lit/ds/symlink/lme49720.pdf


carrotIndustries:
this also changes packages, probably not intended

PR #74, opened 2018-12-04
Added 3D-model for WS2812B

atoav:
Merry Xmas. Decided to model a 3D-model for #59 

One thing: the indent in the datasheet is on pin 3, while in the package the indent is on pin 1 (as usual). How should we deal with this?

See: 
2018-12-04 14_54_10-ws2812b-v4_v1 0_en_18112017214137 pdf - adobe acrobat pro dc  
2018-12-04 14_54_45-window


fruchti:
Thank you! Looks very good to me!

PR #76, opened 2018-12-22
(WIP?) Add LQFP100, STM32F207VGT and variants with less flash

fruchti:
3D model was taken from the KiCad 3D model repository:
https://raw.githubusercontent.com/KiCad/kicad-packages3D/master/Package_QFP.3dshapes/LQFP-100_14x14mm_P0.5mm.step

This can be considered as an example for placing the power pins on top and bottom for experimentation for settling [the power pin placement issue](https://github.com/fruchti/horizon-library-convention/issues/7).

Some open questions for the symbol:

* The text fields for reference designator and value are a bit awkward, but there isn't any other way to place them when all sides of the symbol have pins.
* I chose to place ports A, B and C on the right side and ports D, E and H on the left side of the symbol. Is this counter-intuitive for some?


carrotIndustries:
LGTM, some minor nitpicks

 - the assembly refdes should move to the left edge
 - I prefer having "special" pins such as reset near the top instead of at the bottom

> I chose to place ports A, B and C on the right side and ports D, E and H on the left side of the symbol. Is this counter-intuitive for some?

Any particular reason for placing them this way?

> The text fields for reference designator and value are a bit awkward, but there isn't any other way to place them when all sides of the symbol have pins.

That's the preferred way.

fruchti:
> * the assembly refdes should move to the left edge

I can't find a discussion about the topic right now, but I added this as a rule to the convention for clarification some time ago ([here](https://github.com/fruchti/horizon-pool-convention/commit/7d5a6cb69f1b65edb13162396dc01d8174cb3dfa)). The main advantage is that the position is exactly the same for very small/thin packages, where the “dent” would collide with the $RD text if it were on the left. I agree it looks a bit odd at first for large packages like this one, but we'd have to come up with some boundary for the position switch.

> * I prefer having "special" pins such as reset near the top instead of at the bottom

In this particular case, all pins except reset are usually connected with some passive or directly to ground, so it felt more natural for me to place them near the bottom. But an extra ground symbol doesn't add much clutter, so I have no problem in changing it if you feel different.

> > I chose to place ports A, B and C on the right side and ports D, E and H on the left side of the symbol. Is this counter-intuitive for some?
> 
> Any particular reason for placing them this way?

Because of the usual left-to-right signal flow direction I prefer to have the most-used IOs on the right side (and power supply and special pins on the left). At least for ST microcontrollers, the alphabetically last port most of the times has just a few pins it shares with a crystal oscillator, like port H here. Placing it next to the other special function pins would require them all to be moved to the bottom right side if pins should start with port A on the top left.

Then again, most of the time all IO pins will be connected to labels anyway, so I don't care too much how they are organized, as long as it is somewhat plausible and coherent.

fruchti:
Bot!

fruchti:
With https://github.com/horizon-eda/horizon-pool-convention/issues/7 resolved, I would no longer consider this PR work-in progress.

I fixed some small thins, i.e. the package silkscreen and unnecessary MPN/value duplication in the parts, but left the reset pin at the bottom. Moving it to the top might make adding a resistor to VDD easier in some cases, but would add more confusion if VBAT is used. Keeping it at the bottom feels like the overall better solution to me, as it only requires one extra VDD power symbol in every case.

carrotIndustries:
The assy refdes could be enlarged to fit the assy outline.

One thing regarding the low-active reset pin: Have we yet decided how to do these consistently? Should we still put the dot decoration on the pin even if the pin itself is designated as active low by its name?

fruchti:
> The assy refdes could be enlarged to fit the assy outline.

Will do and add to the convention (which currently says 1 mm maximum). Should we maybe extend the test case `M1234` to a two-letter prefix? We currently have some two-letter ones in the convention and I could imagine some overlap in the packages used for one- and two-letter entities (for example, resistor networks in SO-16 exist).

> One thing regarding the low-active reset pin: Have we yet decided how to do these consistently? Should we still put the dot decoration on the pin even if the pin itself is designated as active low by its name?

I *think* we had that topic already. The convention currently specifies

> 5. Do not use a ‘inverted’ decoration for pins whose name already indicates inversion (n or / in front, overbar, etc.)

which, IMO, follows the spirit of [KiCad’s rule](https://kicad.org/libraries/klc/S4.4/) even though Horizon doesn’t have the overbar decoration.

PR #77, opened 2019-01-08
Circular test pads

fruchti:
I added a few circular test pads.

* Included pad diameters are 0.5 mm, 0.635 mm, 1 mm, 1.27 mm, 1.5 mm, 2 mm and 2.54 mm.
* They are all alternate footprints, since that seemed like the most sensible way of organising them to me. 
* Package and assembly layer are empty. I wasn't able to find a source on how to handle test pad footprints, but keeping those layers empty seemed logical.
* I opted to keep the pad stack “public” so it can be shared between all circular test pads. It could be renamed to something like “Circular SMD pad without paste” to make it more discoverable.


carrotIndustries:
> Package and assembly layer are empty. I wasn't able to find a source on how to handle test pad footprints, but keeping those layers empty seemed logical.

Assembly layer should at least contain the refdes such that test points can be found in assembly drawings.

fruchti:
Good point. These will have to be placed outside the test pad, though, because otherwise “TP1234“ would have to fit into a 0.5 mm circle, making the text unreasonably small.

fruchti:
Added a circle and `$RD` text to the assembly layer for each test pad.

fruchti:
I updated this PR to make some changes to bring it in line with #206 and fix some small remaining issues from that PR:

- Made my alternative symbols reference the unit from #206 and removed the redundant unit.
- Removed my redundant part and entity and made the existing part more generic.
- Added my packages as alternatives to the package added in #206 (removed my redundant 1 mm test point).
- Fixed the file names of #206 not fitting the current convention.
- Made my symbols align with the existing one’s rotation and origin.
- Added the value text to my symbols.
- Switched from quadratic courtyard polygons to circular.
- Harmonised names, tags, and file names.

In contrast to #206, I moved the silkscreen text to next to the test point, which should give a better alignment in average cases (since there no option for centring text). This should, however, the only point of possible breakage, since all UUIDs of the main pool are still in use.

I merged master instead of rebasing because some projects of mine reference UUIDs this update removes and I’d rather not lose that history.

fruchti:
Bot!

fruchti:
Notes on the bot output:

- I aligned the symbols such that the pin connection agrees on all alternative symbols. Naturally, refdes/value aren’t centred around the y axis.
- The silkscreen circles have exactly 0.2 mm clearance to the copper pads. Since the silkscreen is approximated by line segments, these intersect with the keep-out area in some points. The board editor shows this more clearly (2.54 mm test pad as an example):
  2021-04-17-121643_2516x1516_scrot

carrotIndustries:
LGTM, @RX14 your thoughts on this?

fruchti:
TBH, I’m not fully sold on the symbol origin. Having it on the connection point (as it originally was here) seemed to be better UX-wise when placing and rotating test points or exchanging them for alternative symbols, but I didn’t want to break the existing symbol.

RX14:
I'd prefer to move the test point symbol origin to the connection point too, I didn't think of that. If that's okay with @carrotIndustries I'd like to change it now before too many other projects use that symbol.

I'm not really a fan of the silkscreen circle around the test point. The test point copper is always clearly visible against the silkscreen and no assembly is required (unless you use weird silkscreen like the oshpark after dark but thats an edge case)

Everything else is a huge improvement, thank you!

RX14:
Oh, and is there a reason to introduce a new padstack in this PR or would it be better to use the SMD circular padstack from #206

fruchti:
Then I’ll move the origin. Since a cache update of a project is a manual process where you have to expect breakage in any case, I don’t think it’s that big of a deal.

Not sure about the silkscreen circle. I could live without it but I’ve seen it on a lot of board and do quite like the extra contrast. Having both variants would be possible, but might be unnecessary. @carrotIndustries: Your opinion?

> Oh, and is there a reason to introduce a new padstack in this PR or would it be better to use the SMD circular padstack from #206

I had completely missed that, thanks! From what I can tell the padstacks are basically the same. I’ll remove the extra one.

RX14:
Interestingly most of the test points I remember seeing on PCBs have been without circles. I'm not hard against it.

fruchti:
I took the liberty to turn your filled circle test point into a variant of the un-filled one since I’m editing the symbols anyway and that made it a bit easier. I also added a large filled test point for symmetry. If you’d rather have yours back, I can roll that back and stay with your design instead.

fruchti:
Bot!

fruchti:
Hmm, the bot doesn’t seem to display polygon circles as filled like the symbol/schematic editors do.

RX14:
Actually, come to think of it, what is the motivation for all these symbols? Is it not more confusing to have 6 instead of 1 testpoint symbols? What are the usecases?

RX14:
Also, it seems that moving the origin made it really hard to delect the testpoints in the schematic editor...

carrotIndustries:
> Hmm, the bot doesn’t seem to display polygon circles as filled like the symbol/schematic editors do.

Fixed in https://github.com/horizon-eda/horizon/commit/21560ba52fd3369815a4c004ebdf86466a68e789

fruchti:
> Actually, come to think of it, what is the motivation for all these symbols? Is it not more confusing to have 6 instead of 1 testpoint symbols? What are the usecases? Kicad only has a few symbols.

It’s been a while since I opened that PR, but IIRC it was a combination of

- Large/small symbols depending on how much space you want to sacrifice for test points. If you want to add one after the fact, it’s handy to have a smaller symbol.
- Different styles for different types of signals.
- If there is more than one symbol, additional ones don’t add too much since the symbol selection dialogue pops up anyway.

But in truth I think I only used one per project, so keeping only one wouldn’t be too much of a problem for me.

If I could choose only one to keep, I’d probably choose the diamond-shape symbol. These are easy to keep apart from the power symbols with a similar style as the circle-style test point symbols. Filled shapes don’t really appear in schematic apart from net junctions and look kinda odd and out-of-place to me. This is purely a matter of taste, though.

> Also, it seems that moving the origin made it really hard to select the testpoints in the schematic editor...

I just tried it and I think this is because I used a polygon circle for the filled test point instead of arc with larger line width, which doesn’t seem to be included in the bounding box. The non-filled ones seem fine (even though the bounding box for the arcs isn’t optimal, but this is a known issue IIRC). This could be worked around by adding an arc/a line behind the polygon circle.

RX14:
> But in truth I think I only used one per project, so keeping only one wouldn’t be too much of a problem for me.

I think this is an indication we should keep it as-is then. I tried adding small versions of the resistor and capacitor symbol, but I found that it just made me try too hard to keep schematics compact, when in reality the cleanest schematics have large amounts of whitespace and a more thought-out layout.

At the least, one size. Adding different shapes I'll tentatively support, but filled/unfilled seems unnecessary to me.

fruchti:
Ah, yes, I remember that PR. I just looked through my projects and my most-used one is the large diamond test point, so I can definitely get behind removing the small ones. Which one(s) would you want to see in the pool?

I also just noticed that the packages with silkscreen circles could be seen as more generic because you can use ‘omit silkscreen’ in the board editor (and keep the text if you smash the package before it) to get test pads without silkscreen outline. I also did a quick image search and it looks like roughly 70:30 for test point pictures without to with silkscreen outline. You’re probably right in that most PCBs don’t have the outline. Since it’s not that much extra work to remove the silkscreen collectively IMHO, I’d say its worth it to keep the option for both variants.

Another thing I forgot to mention: I named the part ‘Test pad’ with the intention of keeping the current SMT test pads and future THT test points separate parts (instead of having them all alternative footprints for each other). Is that distinction/terminology intuitive to you?

RX14:
> Which one(s) would you want to see in the pool?

I honestly don't have any preference. Circle, filled or not seems to be commonly used by other software. I'd let others with more experience with other software (I've learned mostly on horizon!) weigh in on that.

> use ‘omit silkscreen’ in the board editor (and keep the text if you smash the package before it)

That works, feels a bit hacky and obscure though. If this was added with silkscreen outlines I'd probably end up keeping them.



> Another thing I forgot to mention: I named the part ‘Test pad’ with the intention of keeping the current SMT test pads and future THT test points separate parts (instead of having them all alternative footprints for each other). Is that distinction/terminology intuitive to you?

They should be different parts, yeah. I named the part I added in #206 expecting to add a new part per-package. I'm fine with using alternate packages for different copper test point sizes, but different styles should be a more explicit choice.

fruchti:
Fair points. @carrotIndustries: Opinions regarding schematic symbol style and silkscreen outline?

RX14:
bump @carrotIndustries

carrotIndustries:
I'm not really convinced of the circles surrounding the copper pads, I don't recall having seen that often on other boards. I'd much prefer test points without the silkscreen circle.

Apart from that, the part should set the "no bom" and "no pick&place" flags.

RX14:
@fruchti are you still around to finish this PR? I can take it from here if you'd rather

fruchti:
I removed the silkscreen outlines and set the flags in the part. I wasn’t able to rebase the branch; I’d always end up with conflicts that would be very tedious to solve due to all the moved and renamed files, So, unfortunately, the history isn’t that clean.

@RX14: Thanks for checking on me (and sorry for not being as active as I maybe could)!

fruchti:
Bot!

RX14 (approved):

PR #73, opened 2018-12-03
Added Zener Diode, LM4040, fixed TO-92 3D model

atoav:
This PR turned out to be more than anticipated.

I:
* added a Unit, Symbol and Entity for Zener Diode (I did this by duplicating the Schottky, to stay consistent)
* added multiple LM4040 shunt voltage reference variants both in SOT-23 and TO-92
* added the SOT-23 package by checking it out from my [branch:sot-23](https://github.com/atoav/horizon-pool/tree/sot-23), this is essentially the 3D-model and the package
* added working 3D-models to most TO-92 variants (all of these come from [KiCAD](https://github.com/KiCad/kicad-packages3D/tree/master/Package_TO_SOT_THT.3dshapes)

For the LM4040 I limited myself to Texas Instruments and to the four families:
* LM4040AIZ (TO-92 0.1%)
* LM4040DIZ (TO-92 1%)
* LM4040AIM (SOT-23 0.1%)
* LM4040DIM (SOT-23 1%)

PR #78, opened 2019-01-14
Added Nexperia BZX84-Series Zener Diodes

atoav:
Another big one. I added a zener diode (based on the existing Schottky symbol), I am not sure what the "right" zener symbol is, but I took the one from wikipedia, thios replaces the symbol in #73 

I created a base part and wrote a script to autogenerate the parts from a list of values found in Nexperias datasheet. The part uses SOT-23, which I merged from my branch sot-23.


fruchti:
Some comments:

* The SOT23 3D model seems to be missing
* The tag and folder name “so” is misleading and should be “sot” IMO
* Silkscreen lines do have less than the convention's 0.2 mm clearance to the package body
* Should we enforce packages like “sot23” instead of “sot-23” for the tags, too?

Other than that, it looks pretty good!

fruchti:
For me, SO is synonymous to SOIC. All others should have their own subdirectories. If we want to separate LQFP, TQFP or have them in a QFP directory would be another matter of discussion. Luckily, File names don't really matter for Horizon.

If I remember correctly, I wrote the hyphen-less variant in the convention because of the already existing packages at the time. Personally, I would actually favour the footprint names with a hyphen, too.

I think you should apply the change only to parts from this PR to keep things separated. Other parts could be changed in their PR, or, if they're already merged, in one large “fix to comply with convention” PR once these aspects of the convention have been widely agreed on.

atoav:
Did the proposed move and tag editing for this and the sot-23 branch. I will search for more.

fruchti:
While checking the ‘What is New in IPC-7351C’ slides for something else, I noticed it actually had a drawing for a SOT-23 silkscreen in it:
2021-01-31-142044_4480x1600_scrot

This one seems a bit clearer than the current ‘two lines’ legend to me.

atoav:
@fruchti this is a welcome change, I soldered a few PCBs with the old silkscreen already and this makes things clearer imo

fruchti:
TBH, I also quite like the one Horizon autogenerates from the pads and package polygon. It has lines all around and no pin 1 indication. We could include that one as an alternative package for transistors, diodes, etc. where pin 1 doesn’t really matter (as opposed to ICs).

PR #83, opened 2019-01-18
Add licence based on KiCad's CC-BY-SA exception

fruchti:
Based on http://kicad-pcb.org/libraries/license/, as discussed here: https://github.com/carrotIndustries/horizon-pool/issues/80

[Rendered view](https://github.com/fruchti/horizon-pool/blob/licence-update/LICENSE.md)

To be able to merge, @carrotIndustries and @atoav as authors of parts of the pool would also have to agree on these new terms, and ideally also @m3mory and @greenscreenflicker as authors of currently open PRs. Please also make sure that any third-party content included in your contributions are compatible with this licence.

Suggestions for improvement are welcome!

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,23 @@
+# Horizon Pool License
+
+The Horizon pool is licensed under the [Creative Commons CC-BY-SA 4.0 License](https://creativecommons.org/licenses/by-sa/4.0/legalcode), with the following exception:
+
+---------
+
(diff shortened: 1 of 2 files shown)


atoav:
I am all for it. Sounds good

fruchti:
For the record: I've checked all the third-party files in my PRs and they all comply by having KiCad's exception.

I think we should merge this even if we don't hear from @m3mory and @greenscreenflicker. I'd just leave a comment in their PRs that they would have to agree with the licence before the PRs can be merged.

fruchti:
Thanks! I added a comment to the PRs in question.

PR #84, opened 2019-01-20
Added Generic TH polarized capacitor packages and Panasonic FC-A series

atoav:
The base parts, inheriting parts, packages and 3D-models were generated from data scraped from [panasonics parts list](https://industrial.panasonic.com/ww/products/capacitors/aluminum-capacitors/aluminum-cap-lead/fc?page=1&limit=100), see the Datasheet here: [FC-A Series Datasheet](https://industrial.panasonic.com/cdbs/www-data/pdf/RDF0000/ABA0000C1209.pdf)

I left out parts that only differed in terms of packaging or similar (Lead vs Bulk) by filtering out MPNs that end in M, B or E. From the scraped parts a set of unique package variants with differences in diameter, length, lead spacing and lead diameter have been gathered (35 variants), which I used to generate both the packages and the 3D-models.

Now that I have this working for Panasonics part page I could easily scrape all other electrolytic parts, but I think that would be overkill.


atoav:
I rechecked this one and added parametric data to make it discoverable. This was all generated by scripting (including the 3D models and packages), so I might add more.

Could/Should we add values like endurance and tolerance to the table (more relevant for electrolytics)?

fruchti:
Good work, thanks! Great job especially on the generated 3D models. How did you automate that?

Regarding the names of the packages: Personally, I'd have preferred “height” instead of “length”, since I feel this is less ambiguous. Also, per convention the parameters in the package names should not be in parentheses, not be capitalised and include a space before the unit, e.g. `Polarized capacitor, ⌀ 8 mm, length 15 mm, 5 mm lead spacing`.

Regarding the parametric data, the “type“ column of the parametric table should be “aluminium electrolytic” instead of just “electrolytic“ to be able to distinguish them from tantalum and aluminium polymer caps.

> Could/Should we add values like endurance and tolerance to the table (more relevant for electrolytics)?

IMHO electrolytics should have their own table, since you mostly choose if you wanted a ceramic/film capacitor or an electrolytic first, before looking at actual components.

The electrolytics table should have the additional columns ESR, ripple current rating, tan delta, temperature range, mounting type (axial/radial/SMD). The tolerance is very important for non-polarised capacitors, too IMHO (also temperature drift, which isn't as useful for electrolytics).

atoav:
> Good work, thanks! Great job especially on the generated 3D models. How did you automate that?  

For the measurements I first generate a unique list of packages from all the caps in the the csv, I then use [solidpython](https://github.com/SolidCode/SolidPython) for generating the step models, which is essentially a pythonized variant of OpenSCAD. The cool thing about this, is that it also allows colors. I planned to share the sourcecode for this, but I wanted to clean it up a little beforehand. Right now it is all in a single file (documented for myself), but I think it makes sense to break this one out into a module.

I also got one for WIMA-style PET caps (rounded block with flat bottom and two leads), just in case anybody needs it at some point. Maybe it could make sense to find a generalized system for these modules, and collect them somewhere central, because this could be a great resource for other people.

## Naming/MPN
You are of course right here, I will adapt this soon.

## Parametric Data
Mhh I think a own table could make sense here, since there is a flood of parts, that differ in what they are optimized and there is not much intersection with non-polarized capacitors, other than voltage rating, tolerance and value. I will have a look into this. 

So a table would look like this:
- Value (just like with the non-polarized caps)
- Type (aluminium, aluminium polymer, tantalum, ...)
- Tolerance (in percent)
- Mounting type (axial/radial/smd)
- ESR
- Ripple Current Rating
- Tan Delta
- Temperature Range (maybe this should be split into "min. temperature" and "max. temperature" for better filter experience)
- Endurance (in hours)

How should I call the whole thing? "Polarized Capacitors"?

atoav:
@fruchti @carrotIndustries 
Looks like this at the moment,. I don't know yet how to propperly format the temperature and endurance, but Lukas might tell me soon.


2019-05-13 10_31_36-Pool manager

I think it is beautiful to have filters like that. If I think for a moment about the library experience I have in Eagle, I think I am going to cry tears of joy :' ) You can even select by height which is incredibly useful for stacked PCBs

Ofc there are things we could optimize, E.g. although it is SI comform `mF` is not a very commonly used unit and it could make sense to add a way to leave out certain SI steps (e.g. use `nF` or `μF` instead)

fruchti:
The colour issue has really frustrated me in the past. I think I made the UFQFPN20 3D model in 3 different programs before I had everything working. Good to know that SolidPython can do this!

Is there any reason why the leads of the capacitors are so long?

>     * Mounting type (axial/radial/smd)

Rethinking this, it strictly speaking redundant, since there is also the package column. I still think it is useful for quickly sorting into SMT/THT parts.

> How should I call the whole thing? "Polarized Capacitors"?

I'd have chosen “Electrolytic Capacitors” but that's a matter of taste. Your choice makes just as much sense.

> ## Naming/MPN
> 
> You are of course right here, I will adapt this soon.

The names now have two “mm”…

fruchti:
Also, I just noticed that the packages should probably get a circular courtyard.

atoav:
> Is there any reason why the leads of the capacitors are so long?

I think lazy me just used the numbers from the datasheet, but shortening them makes more sense.  

> Rethinking this, it strictly speaking redundant, since there is also the package column. I still think it is useful for quickly sorting into SMT/THT parts.

I think we should keep it. There are hundred variations of electrolytic packages, which would make quick filtering impractical.

> The names now have two “mm”…  
It was a bit late already, I'll fix it right away.

> Also, I just noticed that the packages should probably get a circular courtyard and assembly and package layers seem to be missing.  

Good catch, when I created the part, circular courtyards where not implemented yet (or if it did I didn't realize they were possible). I will see how I can replace that.

endofexclusive:
@atoav 
Try change the `digits` field? 4 for temperature and 5 for endurance looks all right here.

digits

carrotIndustries:
> Ofc there are things we could optimize, E.g. although it is SI comform mF is not a very commonly used unit and it could make sense to add a way to leave out certain SI steps (e.g. use nF or μF instead)

https://github.com/horizon-eda/horizon/commit/00576c789c65f22261fc25f2b02efff99630b585 adds the option `no_milli`. Values up to 100mF will get displayed as µF, values above 100mF as F.

>  Looks like this at the moment,. I don't know yet how to propperly format the temperature and endurance, but Lukas might tell me soon.

Set `digits` to `-1`.

Apart from that, I'm not sure if it's a good idea to have that much data in there as the columns specified need to apply to everything in that table. For example, height/diameter only makes much sense for radial style packages.

The parametric search also shouldn't be a replacement for digikey or octopart. Anyone looking for a capacitor with special characteristics such as extended temperature range will use one of the mentioned services to find the desired part. From my POV, these columns should be all we need for a generic Electrolytic/Polymer/Tantalum table as we can sure that we'll be able to find a meaningful value for each of these columns for all parts of that table:
 - Value
 - Voltage rating
 - ESR

fruchti:
> Apart from that, I'm not sure if it's a good idea to have that much data in there as the columns specified need to apply to everything in that table. For example, height/diameter only makes much sense for radial style packages.

Should we replace them with generic height/width/length? Along with the definitions that height is always measured upwards from the board and length is the longer side they should be unambiguous and would also work for SMD tantalum caps.

> The parametric search also shouldn't be a replacement for digikey or octopart.

TBH, I thought that was the long-term goal. If that's not it, how is it thought to be used?

niclashoyer:
If we are already scraping manufacturer websites, pull everything out, link to datasheets and filter parts on values, this is essentially the information that octopart holds. So an open database alternative is maybe just the right answer to a commercial product like octopart? If we don't want to bloat the interface in horizon that is reasonable, but we could also include a web interface to the pool, to allow others to access the information we scraped.

niclashoyer:
@atoav CadQuery also looks very nice, thanks for the tip! So it may be possible to automate the footprint generation all the way from scraping to STEP without all the conversions in between.

atoav:
## Height

> Apart from that, I'm not sure if it's a good idea to have that much data in there as the columns specified need to apply to everything in that table. For example, height/diameter only makes much sense for radial style packages.  

Height was meant specificially to be about the vertical extents of the package on the PCB. In the modular synthesizer field I mainly work with stacked PCBs so figuring out which 100μF, 25V cap fits physically with a given clearance (e.g. 1 mm for a PCB spacing of 11mm), is much faster that way than having to figure out the intersection of 100μF 25V caps in the pool with whatever <10 mm I can find on a distributer's site or going and read the datasheet and figure out the right MPN for one that fits and then searching for it.

## Missing Fields/Maybe a solution

I understand your concern when it comes to potential missing data in the future. This is definitly a issue we have to consider in _any_ case (independent of the number of fields). I think a quite elegant solution would be to _allow missing fields_ and add an entry on top of each list like `no data (42)` and when you also select that your filter also includes the 42 parts that have no data on that particular field in the filter. This way missing fields wouldn't be a huge problem because you see they are missing, people who just quickly wanna add any part that fulfills the filters criteria can do so, without worrying about missing fields and people who wanna help with the pool can quickly find parts with missing fields to help. Of course the idea should be to add only fields that apply to all parts.

## Question of Scope

> The parametric search also shouldn't be a replacement for digikey or octopart. Anyone looking for a capacitor with special characteristics such as extended temperature range will use one of the mentioned services to find the desired part.  

Like @fruchti  I must have assumed this was the long term goal. After all we have got this whole "only real parts with real MPNs" paradigma going, so having nice filtering kinda was the logical conclusion in my eyes (it wouldn't make sense if the parts were more generic abstractions). I added that data because it was there and whether I extract one field or 8 made no difference in work. And to be honest, the whole pool concept works so nicely that it really lends itself to something like that kind of approach. My aim here is not to create a "perfectly detailed" parts database for the gist of it – I just want to be able to quickly select a relevant part and not spend more time than necessary for a initial schematic. In the end I will have to go to the distributers anyways and check what the final part will be, at least my workflow always was to export the BOM, make a table out of it, add all the real part URLs calculate the costs etc.

> So an open database alternative is maybe just the right answer to a commercial product like octopart?  

I think that would be an entirely new project with a much wider scope, basically a wikipedia for parts. Would it make sense? Sure! But I would really advice to for now stay focused on horizon itself.

## Future

What is clear to me, is that we certainly need to find some sort of consensus or common vision where the whole pool thing should lead in scope, which workflows it enables, where the traps and pitfalls of messy reality lie, how the whole verification/finding errors part would ideally work out etc. As it is now it certainly works already better than many library concepts I had to deal with, thanks to some pretty good decisions @carrotIndustries took pretty early on, this is a pretty solid basis, and as long as we don't go below this baseline the thing can only become even better.

## Off Topic
As mentioned above I make the whole calculation/parts thing in a seperate table and for this I search the parts and store the links. I know Lukas explicitly said once he doesn't want to have prices and stuff like that in horizon, but even without this it would already be incredibly useful to be able to add custom data to parts on a per-project level, that could be exported with the BOM as a extra custom field. You could store the link to the part there, or the price, or just some note for yourself ("Don't forget M3 mounting screws"). I really _love_ removing little frictions from workflows like that...

niclashoyer:
+1 for the "no data" solution. I think that not all manufacturers will provide data for every field and we may have to deal with that. On to of that it may be useful to select (maybe in global and/or project based configuration dialog) what fields should be displayed, so if "height" was important to me, I could include it. If it is not important for what I'm doing, I may leave it out and focus on other attributes.

niclashoyer:
## Sorry, also OT:
We do have an UUID per part, don't we? Is it possible to export this in the BOM? I already have a calc table with all my parts and Reichelt/Mouser SKUs. Then I could just add the UUIDs and match from there.

carrotIndustries:
Ok, I'll ad None/NULL values for missing fields.

Aren't dimensions (width/height) parameters of the package rather than the individual parts? This got me thinking if adding parameteric data to packages as well would be a sensible idea.

Instead of specifying fixed columns as with the part's I've had more of an ad-hoc approach similar to packages in mind. Package names such as `Pin header, 10 pins, 2 rows, 2.54 mm pitch` are already a serialized form of parametric data, so why not make it a first class citizen? This data would then be joined into the parametric parts listing.

Regarding which columns to have: If we want to have width/height, we should probably have that for all tables? Digikey at least specifies "Lifetime @ Temperature" and "Ripple current @ frequency", so just having a single figure doesn't appear to be that useful to me. We should probably have look at some other series before settling on the columns.

Re scope: My opinion on not trying to replace octopart/digikey was to limit the scope in order not get sidetracked in trying to categorize everything the industry has to offer.

>  to add custom data to parts on a per-project level, that could be exported with the BOM as a extra custom field. You could store the link to the part there, or the price, or just some note for yourself ("Don't forget M3 mounting screws"). I really love removing little frictions from workflows like that...

I've though about that as well, but once again the the difficult thing is to get the scope right so that we don't get sidetracked implementing a poor spreadsheet application.

atoav:
> Aren't dimensions (width/height) parameters of the package rather than the individual parts? This got me thinking if adding parameteric data to packages as well would be a sensible idea.  

Okay now that is a cool idea. Parametric data in packages makes kinda sense, given the direction our current solutions are going. I am not sure however if using the filename to generate that data is really the best aproach, because it could raise a whole new class of problems with user input and validation. I'd rather have it similar to the way you implemented it with the parts and then in the package editor there is dialog to enter that stuff.  

I am not sure if width/height makes sense for all parts, for resistors for example the package (e.g. 0805) is usually enough to know the size, unless we are speaking of things like high power resistors. Things like width/length can easily be derived from the package outline and height can be entered manually.

> Digikey at least specifies "Lifetime @ Temperature" and "Ripple current @ frequency", so just having a single figure doesn't appear to be that useful to me.  

I wondered a little about this one too, but my feeling was that the filter is just for rough narrowing down (so 1000h vs 10000h), anybody who really needs to know the durability of a part with high confidence just _will_ have to read the datasheet anyways. I am not to sure how this goes with ripple current @ frequency tho, maybe it makes sense to be more accurate here?

> Re scope: My opinion on not trying to replace octopart/digikey was to limit the scope in order not get sidetracked in trying to categorize everything the industry has to offer.  
  
I am on your side there, as nice as it could be for this to become a fully fledged thing that can do _everything,_ it is arguably better to focus on the core functionality. Which is drawing schematics and boards, with a nice and sane library managment, that doesn't make you tear your hair out. IMO decent filtering should be part of that, but always with the goal in mind: it is about quickly checking if you find a fitting part, not about having a perfect compendium of all human artifacts that could find their way on PCBs. The filtering that is possible with the parametric data is just too nice to throw away in regards of horizons core functionality, it really _does_ make a difference.

Arguably the parametric data is not as important for ICs and similar things. Usually there aren't that many and you know them more or less by name anyways, so tags might be enough here.

## Spreadsheets

> I've though about that as well, but once again the the difficult thing is to get the scope right so that we don't get sidetracked implementing a poor spreadsheet application.

This is also something I would avoid. Making it possible for users to attach custom data/notes to the parts on their schematic (and making them available in CSV export) or allow them to export UUIDs like @niclashoyer mentioned wouldn't really amount to any spreadsheet-like capability in my eyes.   

I could imagine something that just links to a Parts UUID (if the UUID is not found it just gets ignored) like this:  

pool_diagramm_full_1080

And then on BOM export you can select these fields to show up or something.

The main reason why I'd argue for sth. like that is that it allows users to keep data with the parts instead of having to manualy update some spreadsheets, when they export a new and changed BOM.

fruchti:
I thought about the parametric data in general for a while and I want to revise some of my prior points.

> Re scope: My opinion on not trying to replace octopart/digikey was to limit the scope in order not get sidetracked in trying to categorize everything the industry has to offer.

In my opinion, the exact scope is the first thing we should define before we carry on with the discussion about individual features. The features should depend on the use case, not the other way round.

I think if parametric data didn't already exist, I would have voted against including it into Horizon. I think that a small-footprint solution (what currently exists) is of limited usefulness compared to the extra work needed to curate the pool. Currently, you can search for passives but for nothing else, a distinction that is pretty arbitrary IMHO. For the majority of part choices for a new design, you'd use the distributors'/manufacturers' parametric searches anyway. Perhaps I'm mistaken here, but I don't really see a benefits for a minimal implementation.

A fully-fledged implementation of a parametric search, on the other hand, might be very comfortable for the end user, but it disagrees with key design decisions of the pool. The pool does not have categories and is organised in a flat way. With parametric tables, we're re-introducing categories (and for a fully featured parametric search, these would have to be nested) and we would end up with a pretty confusing pool structure, as the directory structure possibly doesn't match the parametric categories.

Is removal of the parametric search altogether an option? I know it sounds harsh and I don't want to depreciate the work you've put into this. I just have the fear that this is the start of feature creep and limited use cases are included into Horizon that are handled way better by existing or potential external applications.

I'm absolutely convincible if you have a use case in mind for a parametric search fitting into Horizon's concept. So far, I have doubts this feature is really worth supporting.



> Aren't dimensions (width/height) parameters of the package rather than the individual parts? This got me thinking if adding parameteric data to packages as well would be a sensible idea.
> 
> Instead of specifying fixed columns as with the part's I've had more of an ad-hoc approach similar to packages in mind. Package names such as `Pin header, 10 pins, 2 rows, 2.54 mm pitch` are already a serialized form of parametric data, so why not make it a first class citizen? This data would then be joined into the parametric parts listing.

At first sight, this is very appealing. The now-existing conventions about the parameters would be formalised and fixed. The data would be easily usable for search and export. The only way to implement this without a notion of categories would be to offer all possible columns for each part, though.

If we restrict them to mechanical properties (height, width, depth, diameter, pin pitch, body width, mounting angle, …) and some booleans (hand soldering, socketed, …), it might be feasible without getting too crowded.

Yet either all distinctions needed for a single part would have to be added as columns or there would be some “extra notes“ field with a function similar to the existing name. Both solutions don't sound too great to me.

carrotIndustries:
> Currently, you can search for passives but for nothing else, a distinction that is pretty arbitrary IMHO.

That's the primary raison d'être for this feature. I've found passives to be one of the most annoying parts to deal with in schematic design. Given that the pool contains more or less all of the generic passives you need in your design, how to search for these knowing that you'll need an X7R 68nF capacitor in 0603 package? Adding parametric data to the parts was my solution to this. In some (albeit not at all useful) way, this has been part of horizon and the pool since day 1: https://github.com/horizon-eda/horizon/blob/429222a46c13928b8b6da8323c0378c7b0a2028f/pool/part.hpp#L47

I've also got plans on making it possible to make the parametric values show up in the schematics to better help communicate the design intent.

It's not my intention to extend parametric data to more than jellybean passives as the other parts are discoverable by tags and search.

> Yet either all distinctions needed for a single part would have to be added as columns or there would be some “extra notes“ field with a function similar to the existing name. Both solutions don't sound too great to me.

Any other suggestions? The KLC mandates what's more or less a poorly serialized implementation of parametric data: http://kicad-pcb.org/libraries/klc/F2.2/ and ends up with hard-to-parse names such as `SOIC-8_3.2x5.7mm_P1.27mm`

fruchti:
> That's the primary raison d'être for this feature. I've found passives to be one of the most annoying parts to deal with in schematic design. Given that the pool contains more or less all of the generic passives you need in your design, how to search for these knowing that you'll need an X7R 68nF capacitor in 0603 package?

Okay, I see. If the primary use case is mostly generic parts where the exact characteristics don't matter too much, this should be clarified. So, power inductors, as well as specific properties as electrolytic ESR would then already be out of scope? 

> this has been part of horizon and the pool since day 1: https://github.com/horizon-eda/horizon/blob/429222a46c13928b8b6da8323c0378c7b0a2028f/pool/part.hpp#L47

Yes, I've noticed that. The generic JSON field was too abstract for me to think about its possible uses, though.

> The KLC mandates what's more or less a poorly serialized implementation of parametric data: http://kicad-pcb.org/libraries/klc/F2.2/ and ends up with hard-to-parse names such as `SOIC-8_3.2x5.7mm_P1.27mm`

The KLC was the primary reason for me to propose the explicit and verbose naming now in the convention. I guess the display in the part selector could stay more or less the same even if there were columns for parametric data. Empty columns would be invisible and hiding columns or sorting based on them would still be possible.

atoav:
> In my opinion, the exact scope is the first thing we should define before we carry on with the discussion about individual features.   

Agreed. I think however that for communication it is sometimes necessary to flesh out your "vision" of how things could be and there it could be usefully to get concrete and not hide potential problems with your idea behind the curtain of abstraction. So it might be necessary at times to make solutions or features tangible to get a clearer idea what this would really mean in practice. The things we _really_ should avoid is bike-shedding about minor details that won't really matter that much in the end or ideas that are so much out of scope that all they do is distraction.

> I just have the fear that this is the start of feature creep and limited use cases are included into Horizon that are handled way better by existing or potential external applications.

I understand the point you are making here. Every developer knows the danger of feature creep and that is why I usually try to stick to the usage: if a feature helps the usage in a good way, without introducing tons of complexity, it is good. And reducing complexity is also good. But reduction of complexity is not _always_ a desireable goal – in _certain_ use cases you can't reduce complexity, you can only offload it to the user.

When a user wants to draw a circuit, the first thing they would usually do is to search for parts that are already there (at least in the case of more or less generic parts like electrolytics, capacitors, resistors and inductors like Lukas also said). Creating a new part is work and has to be learned. This means: better filtering helps users to make the decision whether they need to create new parts or not, is especially valuable to new and inexperienced users and raises the value of the whole pool, if the scope doesn't get out of hand. 

> It's not my intention to extend parametric data to more than jellybean passives as the other parts are discoverable by tags and search.

I agree that this is most useful for LCR parts and I would be perfectly fine with limiting it's use to that.  

Here a quick pool illustration to lighten the mood:  
Untitled-1-01

Also: Ultimately let's not forget that the cool thing here is, that the tables are part of the pool, so if anybody should decide to have a need for it, they could easily go in and use it in a different way : )

niclashoyer:
Sorry to dig this up. Is there any chance to get this (and other generic parts) forward? Now that 1.0 is out it might be a good idea to have more generic parts ready? I'm constantly pulling in generic parts from different repositories into my own fork.

atoav:
I think users could benifit from a couple of common parts in the offical pool, I am willing to adjust this  to whatever just to get it accepted – any ideas how to move forward on this one @carrotIndustries and/or @fruchti ?

frmdstryr:
Any chance this is in a separate pool somewhere? Or could it be moved to one (like you did for [the EEHZA caps](https://github.com/atoav/horizon-pool-panasonic-eehza-capacitors))?

frmdstryr:
Made one from this branch [horizon-pool-panasonic-fc-a-capacitors.zip](https://github.com/user-attachments/files/18462154/horizon-pool-panasonic-fc-a-capacitors.zip)

PR #85, opened 2019-01-20
Added some 3D packages for the DIP packages

atoav:
All models come from [kiCAD](https://github.com/KiCad/kicad-packages3D/tree/master/Package_DIP.3dshapes). I combined the sockets and the packages in case of the socket models


atoav:
The pin spacing has been fixed here and in all other of my PRs containing dip packages.

PR #93, opened 2019-03-25
Added OPAx197 by TI

atoav:
I added the OPAx197 series by TI as outlined in [this datasheet](http://www.ti.com/lit/ds/symlink/opa4197.pdf).

I left the VSSOP variants out, because I was lazy. If anybody has a package for that, feel free to add.


carrotIndustries:
Bot!

PR #94, opened 2019-04-03
Added LF398 Sample and Hold IC

atoav:
I added the SOIC-14 variants of the LF398 Sample-and-Hold IC: http://www.ti.com/lit/ds/symlink/lf398-n.pdf


carrotIndustries:
Bot!

PR #95, opened 2019-04-20
ADS1220

krasimirk:
Hello,
I would like to add ADS1220 4-channel 24-bit ADC from TI into horizon pool.
Kind Regards,
Krasimir Kostadinov

### DIFF
--- a/entities/ic/adc/ti/ADS1220.json
+++ b/entities/ic/adc/ti/ADS1220.json
@@ -1,0 +1,19 @@
+{
+    "gates": {
+        "67684ec3-0a8c-4b57-9195-b369cbcdedde": {
+            "name": "ADS1220",
+            "suffix": "",
+            "swap_group": 0,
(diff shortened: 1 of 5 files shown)


carrotIndustries:
Symbol is all over the place:
 - refdes/value are placed in the middle of nowhere
 - avdd/dvdd pins are upside down
 - vertical pin labels should be perpendicular to pin
 - pins should be 2.5mm (2 grids) apart
 - use alt. names for alternate pin functions

Misc:
- Seems like TI's RVA package doesn't quite match our QFN-16 package, please also add the RVA package according to their drawing
 - Gate should be named `Main`
 - Entity should have prefix `U`

krasimirk:
Closed pull request, since after correction I made a new one.
The new one have description "Fixed remarks from Lukas K.". 
Sorry, that the name does not nothing to do with the ADS1220. I miss to add it. 
I though that the new pull request will be added into the current one.

PR #96, opened 2019-05-01
nRF52811

cortesis:
Created a new Padstack with both ends rounded.
Created package QFN-48 for nRF52811 (same measures as on Altium Nexus)
Added 3D-model for QFN-48 from KiCad
Created unit,entity,symbol and parts for all nRF52811-QFxx variants


carrotIndustries:
- Use base part for avoiding duplicating pin/pad assignment
 - Symbol shouldn't look like the package, organize pins by function (have a look at other symbols in the pool)
 - Ground pad should be named PAD
 - Package should either be generic or manufacturer-specific. For generic QFNs, use d-shaped pads.
 - 3D model is for 40 QFN

PR #99, opened 2019-05-01
Added NX3225SA-32M-EXS00A-02994

cortesis:
Created NX3225SA-32M-EXS00A-02994 from base part NX3225SA


cortesis:
Duplicate of request #100

PR #105, opened 2019-05-15
Add SOT-223 footprint

frmdstryr:
3D model is from KiCad


jue89:
I think this can be closed. This commit has become part of the Horizon Pool with #162, after a little adjustment requested by @carrotIndustries. Thank you @frmdstryr for sharing your footprint!

PR #107, opened 2019-05-19
Added Intel Max 10 10M25SCE144C8G

endofexclusive:
NOTE: 10M25SCE144C8G is in a different pin migration set than 10M50SCE144C8G and 10M40SCE144C8G.


carrotIndustries:
Is the empty space between the pins intentional?

2019-05-19-234151_505x345_scrot

endofexclusive:
Oops. The name of the symbol was also wrong (forgot to rename after copying from the 10M50 part). So that made me miss updating Bank 4 when going through them in the pool. Should be fixed now.

PR #109, opened 2019-05-28
Please, add ADS1220

krasimirk:
1. Refdes/value are placed in the middle of nowhere
-> Placed near the symbol, from the right side as many other symbols

2. avdd/dvdd pins are upside down
-> fixed

3. Vertical pin labels should be perpendicular to pin
-> fixed

4. Pins should be 2.5mm (2 grids) apart
-> fixed only for the horizontal pins, skipped for vertical ones, because there is no space due [3]

5. Use alt. names for alternate pin functions
-> fixed

6. Seems like TI's RVA package doesn't quite match our QFN-16 package, please also add the RVA package according to their drawing
-> used land pattern data from the pre-last page of the datasheet to fix the pad size and the space from the heat pad and pads.
-> not sure if the courtyard and assembly polygones are in the correct sizes

7. Gate should be named Main
-> fixed

8. Entity should have prefix U
-> fixed

### DIFF
--- a/entities/ic/adc/ti/ADS1220.json
+++ b/entities/ic/adc/ti/ADS1220.json
@@ -1,0 +1,19 @@
+{
+    "gates": {
+        "67684ec3-0a8c-4b57-9195-b369cbcdedde": {
+            "name": "Main",
+            "suffix": "",
+            "swap_group": 0,
+            "unit": "92d02b88-f6d6-4f44-827d-24202e294028"
+        }
+    },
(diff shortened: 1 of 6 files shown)


carrotIndustries:
Bot!

carrotIndustries:
- Symbols need clean up to make checks pass
 - Connect pad to GND
 - Parameter program looks strange
 - Package isn't generic, remove tag
 - Why is assy outline larger than package?

PR #111, opened 2019-06-20
Add TPIC6C595, SO-16

fruchti:
This PR contains the TPIC6C595 power shift register, a SOIC-16 package and KiCad's 3D model for it. Depends on [the power pin placement issue](https://github.com/horizon-eda/horizon-pool-convention/issues/7).

I'm also not sure if the open drain symbols on the output pins are really necessary, since the pin names are already “DRAINx”.

The part is already in use on a PCB:
so16


fruchti:
Updated, now uses only a single part with four orderable MPNs instead of four parts.

fruchti:
Bot!

fruchti:
Since https://github.com/horizon-eda/horizon-pool/pull/11#issuecomment-759699745 is now resolved, this PR is no longer blocked.

However, the bot still complains about the tag ‘shift register’. Is there a reason to disallow tags with spaces? It seems to be handled fine everywhere in the UI.

carrotIndustries:
> It seems to be handled fine everywhere in the UI.

Tags show up space-separated in treeviews (though this could be changed to commas).

fruchti:
> > It seems to be handled fine everywhere in the UI.
> 
> Tags show up space-separated in treeviews (though this could be changed to commas).

Good point. Personally, I’d rather have them comma-separated than have to replace spaces in tags (e.g. `shift_register`). Then again, limiting tags to a single word might be a good rule to prevent overly-specific tags.

I could see a tag like `shift register` for this part as useful for finding parts in the pool. If this isn’t too fine-grained for you, I’d add it to the part in the PR as well.

PR #113, opened 2019-06-25
Add LQFP-100 package and 3d model

frmdstryr:
3D model is from KiCAD


frmdstryr:
Didn't see this is already in #76

PR #114, opened 2019-09-05
Added LM13700 OTA and SOIC-16 package

atoav:
The SOIC-16 Package is based on the existing soic-14 package. The 3D-model is taken from the KiCAD collection.

The LM13700 is a operational transconductance amplifier (OTA) by TI. It is based on [this datasheet](http://www.ti.com/lit/ds/symlink/lm13700.pdf). The LM13700 is definitly the most famous and widespread OTA IC, but includes a darlington buffer for each channel. I decided to split these off into their own unit, because that is also how they are often used. The collectors of these darlington buffers are internally connected to `Vcc`. The best solution in my eyes was to signal this using a "faux" power symbol in the Darlington Buffer Symbol. The symbol of the darlington buffer is copied from the NPN symbol.


atoav:
I realized I already made a soic-16 package in #68 so I used that one instead.

carrotIndustries:
Bot!

carrotIndustries:
- MX/M should be orderable MPNs
 - open collector symbol on a transistor isn't really necessary
 - overlapping text in OTA symbol, text placements aren't needed for these kind of symbols

RX14:
Also, both check failures on the SOIC-16 are valid and need fixed. Courtyard generator and silkscreen generator with their default 0.2mm parameters should help you do that for you. The number of silkscreen lines is correct though.

This is 90% there though! And a good job on the units and entity.

atoav:
@carrotIndustries @RX14 Thanks for having a look

Not sure whether the warnings `warning Pin + has improper orientation` and `warning Pin - has improper orientation` are something that demands action, but other than that I would be fine with this.

I used this part on a working PCB as well, so the footprint and the mapping is correct and tested

RX14:
The `LM13700` part should be renamed `LM13700M`, and `LM13700N` should have it's value cleared (ICs don't need a value).

I would also move the part/entity/unit/symbol files under `ic/opamp/ti`.

The package silkscreen outline should have the same number of lines as the old version, i.e. remove the lines between the pins, and there should be only one line at the bottom. I hope you get the gist. Basically the rules are [these](https://klc.kicad.org/footprint/f5/f5.1/) except standardizing 0.2mm space between package outline and silk.

atoav:
> The LM13700 part should be renamed LM13700M, and LM13700N should have it's value cleared (ICs don't need a value)

Good catch

> I think removing the soic-16 tag would be good too, there's no need for the package in tags when the package is part of the part.

This is something I have done on nearly every IC I created since 2018, and it is also something I commonly use to filter things, so I'd rather not do this, sorry.

>  Basically the rules are these except standardizing 0.2mm space between package outline and silk.

This part was just created _before_ the `Generate Silkscreen`-Tool existed, the 0.2mm convention already existed, but no automatic rule checks for silkscreen distance, which is why this one slipped. Because the `Generate Silkscreen`-Tool created it that way and I was out of the discussion for a while, so I assumed the convention might have changed and accpeted the result the machine god gave me. Anyways, I fixed it.

> I would also move the part/entity/unit/symbol files under ic/opamp/ti.

I moved it to `ic/manufacturer/ti` instead, as this is arguably not an opamp and TI are the only ones who make it. If I remember correctly Lukas intended `ic/manufacturer` for that kind of manufacturer-specific parts. I think the folder structure is not that strict anyways, but removing spaces from the manufacturer name makes a ton of sense afaik.

> The warnings are fine, the symbol warnings are fairly useless for non-rectangular symbols. And these symbols are, quite honestly, fantastic.

Thank you very much. I remember having quite a fight with the circles back then : )

atoav:
Bot!

atoav:
Bot!

RX14:
> which is why this one slipped

The rest of the SOIC packages need updating to this convention too. I'll go over a bunch of common packages and update them to be consistent and named one day, but that's another PR...

RX14:
symbol/unit/entity paths need updating too. The convention is to name after the part number, so no `_soic-16_base` or `_darlington_*` please.

atoav:
> symbol/unit/entity paths need updating too. The convention is to name after the part number, so no _soic-16_base or _darlington_* please.

The `_soic-16_base` suffix was the convention at the time the part was created (and encouraged by the creator of the software), because orderable parts didn't exist back then –  this is not needed anymore in this case, so it can go. I am all for conventions, but I will break them where they get silly, and IMO calling a symbol `lm13700` while it is the symbol for the internal darlington pair of a LM13700 is just that. There is a saying by famous artist Nam June Paik when he was asked why one of the televisions he ordered for his huge TV installation was missing: _When too perfect, Gott böse_

As Lukas mentioned multiple times, these paths are not really crucial to get 100% perfect: Otherwise we would have decided to come up with a clear structure and categories, which we decidedly didn't, because we have a pool with objects identified by their UUIDs. This means the structure on disc is mainly for the comfort of people using git by hand (like me), but not for any critical aspects that could spin out of control (like in a typical EDA library). So getting the paths perfect is nice to have but not strictly needed – and quite frankly I won't have the time to do this for every part I submitted, although I _get_ the appeal of it. Keep in mind, that a lot of these parts came from a time, where rule checks weren't there like they are now, the Bot didn't exist and a lot of directories didn't exist as well, so creating a part meant coming up with something.

That being said I made it look nicer now and changed the directory names to `ti`.

atoav:
Bot!

RX14:
> IMO calling a symbol `lm13700` while it is the symbol for the internal darlington pair of a LM13700 is just that

Sorry for the misscommunication! I meant just the `s/ti/texas instruments` on those ones.

Didn't mean to nag so much about the names, they're perfect now!

RX14:
Looks all good to me now.

atoav:
> Didn't mean to nag so much about the names, they're perfect now! And I'll try to keep my perfectionism in check in the future :)

Hey, no problem. It is very good to have somebody who looks at these things with a keen eye (which you certainly have). Just make sure to signal clearly which steps are crucial, and which ones are just "nice to have" in order to avoid scaring away contributers that aren't blessed with patience or don't share your perfectionism.

Of course if the other person won't do it (or it seems a bit much, because it is just a minor nitpick), there is always the option to propose changes and offer to do the commit for them. Lukas changed some things on my PRs as well and I was always happy with the things he did : )

RX14:
Yeah, I'm not entirely sure I have the access to push commits to other's PRs...

atoav:
> Yeah, I'm not entirely sure I have the access to push commits to other's PRs...

Ah.. right not sure about that.

carrotIndustries:
> parts/ic/manufacturer/ti/lm13700n.json

I think it's better off in `parts/ic/opamp/ti/`

atoav:
Bot!

atoav:
Bot!

atoav:
@carrotIndustries Okay this is kinda ready but it seems the conflict with `packages/ic/smd/soic/so-16/package.json` throws the bot off (silkscreen too close to package), I am unauthorized to resolve the conflict, but as it just makes the silkscreen from pre-bot-times conform with the current rules I guess it is something we want anyways.

The warnings for the symbols can be ignored AFAIK.

RX14:
What changed since the last review, and what's up with the unmapped pins?

atoav:
> What changed since the last review, and what's up with the unmapped pins?

I changed the soic-16 silkscreen. However, because soic-16 is already in master, there is a merge conflict and because there is a merge conflict (I suspect) the pot is unable to check the mappings. The editor shows the mapping locally however.

RX14:
This is also important because the soic-16 part needs to keep the same UUID.

atoav:
> This is also important because the soic-16 part needs to keep the same UUID.

They both have `ae1c4d81-1bcd-4e4b-9ec7-8f5c988c6918` so no changes there.

> If you rebase to solve the merge conflict, there shouldn't be a problem?

I am not entirely sure how to do that, any pointers?

RX14:
> I am not entirely sure how to do that, any pointers?

you have to do it using a git client, so check out horizon-pool and add your fork as a remote, then switch to the lm13700 and `git rebase origin/master`. It'd be really useful to have this into the pool PR manager though.

atoav:
Bot!

atoav:
Okay I learned something about rebasing now, worked pretty well, and we have only one commit now : )
Unless someone still finds something this would be ready : )

RX14:
LGTM

PR #116, opened 2019-09-11
Added 2N390x transistors

atoav:
Added the through-hole transistors by onsemi:
- 2N3904
- 2N3904G
- 2N3906
- 2N3906G

Datasheets:
https://www.onsemi.com/pub/Collateral/2N3906-D.PDF
https://www.onsemi.com/pub/Collateral/2N3903-D.PDF

### DIFF
--- a/parts/transistor/2N390x/2N3904.json
+++ b/parts/transistor/2N390x/2N3904.json
@@ -1,0 +1,30 @@
+{
+    "MPN": [
+        false,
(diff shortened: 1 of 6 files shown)


carrotIndustries:
Datasheet URL spilled over into manufacturer field.

atoav:
Good catch! Fixed in latest commit : )

PR #117, opened 2019-09-15
Stackpole & muRata parts download scripts

eqvinox:


carrotIndustries:
Great stuff! Some minor details still need work before we can merge this:

Add base parts for each series you're generating. If there's one datasheet for all parts in the series, add it to the base part.

Murata capacitors don't have a datasheet

Unfortunately, the MPNs listed in the murata source omit the last letter specifying the packaging type. Some seem to be available in D and J whereas others are only available as one of them even if the murata website lists both such as 	GRM31A5C3A100JW01J. We need the full MPN since the upcoming partinfo API integration will require these for exact matching. My proposal thus is to add the D and J suffix for all GRM capacitors in the orderable MPN field. If one of them isn't available at all, it's not too bad as the partinfo API integration will check for all orderable MPNs.

eqvinox:
A-yup I'll go fix this stuff, will probably find time in a few days or so.

Looking forward to that partinfo API :grin:

eqvinox:
> My proposal thus is to add the D and J suffix for all GRM capacitors in the orderable MPN field. If one of them isn't available at all, it's not too bad as the partinfo API integration will check for all orderable MPNs.

This turned out to be insufficient since it's only `D,J` for some, but `D,J,W` and `L,K` occur too.

I'm now scraping the HTML product info page for each `(series, package, height)` combo once and using that (really don't want to download every single HTML product page...)

carrotIndustries:
Just noticed that the same is true for the stackpole ones:

There's RMCF0603F**G**1K00 and RMCF0603F**T**1K00. Both are actually the same part. The only difference is reel size.

eqvinox:
should be fixed, sorry for the delay, was a bit busy

carrotIndustries:
Orderable MPNs LGTM. description should use `Ω` rather than `ohm`

PR #118, opened 2019-09-15
parts: Stackpole RMCF & muRata GRM autogenerated

eqvinox:
Separate from PR #117 to make reviewing a slight bit easier.

Also, "update pool" is now slow as a snail :crying_cat_face:


carrotIndustries:
This seems to include parts not generated by the current version of the scripts such as GRM188R70J104KA01.

eqvinox:
> This seems to include parts not generated by the current version of the scripts such as GRM188R70J104KA01.

Interesting. I haven't created any parts manually, but I did run the muRata script against the web search & s2p model files before (instead of using the "download all" CSV.)  I really hope those lists are not inconsistent with each other :open_mouth:

I'll go look into it, it's probably gonna be a few days though.  Thanks for pointing this out!

eqvinox:
> I'll go look into it, it's probably gonna be a few days though. Thanks for pointing this out!

Seems the simulation models included specs for obsolete parts.  I've removed everything and regenerated it.

PR #120, opened 2019-09-25
DEVKIT-MPC5744P_NXP

Peta-T:
Added DEVKIT-MPC5744P_NXP


carrotIndustries:
- symbol is missing
-too long pin names

Nevertheless, I think that plug-on boards like these don't belong in the pool.

PR #123, opened 2019-10-19
Added Frames with license field and OSHW logo

atoav:
I added two frames based on the existing A4 frame adding a `$license` field for open hardware projects. I also added a variant including the OSHW Logo for Open Hardware projects

### DIFF
--- a/frames/a3_landscape_license.json
+++ b/frames/a3_landscape_license.json
@@ -1,0 +1,421 @@
+{
+    "arcs": {},
+    "height": 297000000,
+    "junctions": {
+        "007617e5-2d7f-4390-bd63-22d149e6e4df": {
+            "position": [
+                341750000,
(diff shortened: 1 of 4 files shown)


carrotIndustries:
Everything in schematics and frames should go on layer `0`. (The logo is on layer `10000`)

atoav:
Okay weird, I didn't select any different layer. This might have to do with the `import dxf` command – could it be that it doesn't import to the default layer?

carrotIndustries:
>Okay weird, I didn't select any different layer. This might have to do with the import dxf command – could it be that it doesn't import to the default layer in the Frame editor?

fixed in https://github.com/horizon-eda/horizon/commit/4f49f9d535634110c9a25e526eb1a14dc5ab8f0a

PR #124, opened 2019-10-19
Added Generic 9v battery clip

atoav:
This is a bid of an oddball, because these cabled 9V battery clips rarely have a well defined MPN (but because they are cabled that rarely matters). I am not sure how we should deal with parts like these, but they are quite useful for private usage, so including these won't hurt I thought.


atoav:
This is probably not relevant for this pool, I will move it somewhere else

PR #125, opened 2019-10-19
Added Vishay kseries ceramic through hole capacitors

atoav:
To end the distinct lack of th ceramics I added the K-Series by Vishay. Datasheet is [here](http://www.vishay.com/docs/45171/kseries.pdf).

A few things to note:
- I included the generator scripts
- 3D models were generated by script
- I skipped some things (e.g. there are four different lead shape variations that don't impact the footprint), I decided to just do the straight ones – everything else would have been overkill


carrotIndustries:
- perhaps make the solidpython dependency optional so that stock python is sufficient to generate the parts
 - the preferred way to get reproducible UUIDs is to use the `UUIDGenerator` from `util.py`

PR #126, opened 2019-10-21
Added 2N7000 Small Signal N-Channel Mosfet

atoav:
Added the [2N7000](https://www.onsemi.com/pub/Collateral/2N7000-D.PDF) by On Semi.
The MPNs are a bit confusing, because the Datasheet is for the 2N7000G, but I only found the part under different MPNs as buyable. But I think it is ok.

### DIFF
--- a/parts/transistor/nmos/2N7000/2N7000 (Base).json
+++ b/parts/transistor/nmos/2N7000/2N7000 (Base).json
@@ -1,0 +1,49 @@
+{
+    "MPN": [
+        false,
+        "2N7000G (Base)"
+    ],
+    "datasheet": [
+        false,
+        "https://www.onsemi.com/pub/Collateral/2N7000-D.PDF"
+    ],
+    "description": [
+        false,
+        "Small Signal N-Channel MOSFET 200 mAs, 60V"
+    ],
+    "entity": "c135130c-0663-4575-9083-3b328ac00c90",
+    "inherit_model": false,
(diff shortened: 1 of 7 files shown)


carrotIndustries:
In that datasheet https://www.onsemi.com/pub/Collateral/NDS7002A-D.PDF some of the MPNs you mentioned only differ by packing method. Instead of distinct parts, create one part and add these as orderable MPNs since one will receive exactly the same thing when ordering a 2N7000-D74Z or a 2N7000.

atoav:
Okay fixed these issues. Decided to also update the existing SMD variant of the part by:
- Adding some tags to the base part (it only had the `mosfet` tag, I added `smd`, `n-channel` and `sot-523`)
- Renaming the 2N7002-7-F to 2N7002 and adding all the orderable MPNs as orderable MPNs

carrotIndustries:
Bot!

atoav:
Bot!

atoav:
Weird, the value on 2N7000 *is* blank. I even checked manually in a text editor before pushing. Or am I getting something wrong here?

carrotIndustries:
The 2N7000 inherits from 2N7000G (Base) that has the value set to 2N7000, hence the warning.

atoav:
Bot!

carrotIndustries:
Bot!

atoav:
Bot!

PR #129, opened 2019-11-02
[WIP?] Base 0603 solder jumper

fruchti:
I wanted a solder jumper footprint where optionally a 0 Ω resistor could be placed. A standard 0603 footprint is usable as a solder jumper, but shows room for improvement. This design is based on the footprints ST uses for their Discovery boards (both as user-solderable jumpers and for 0 Ω resistors) and Horizon's default R0603 footprint. I included two alternate packages, with and without paste.

Note that the ‘value’ field of the part deliberately contains a single space. This way, the part can be used in a schematic without having to smash it to remove the otherwise annoying value.

I marked this PR as work in progress because I lack the experience to judge if my footprint will actually work out in a reflow process in practice. I haven't really found information or examples on this subject, so if anyone has some pointers I'd be very grateful.


fruchti:
I added a keepout polygon to prevent filled planes ‘leaking’ into the gap between pads.

fruchti:
> I marked this PR as work in progress because I lack the experience to judge if my footprint will actually work out in a reflow process in practice.

I have since then used this footprint with a 0.15 mm stencil and no paste mask contraction in [a project](https://github.com/fruchti/leshy). I had no problems during reflow and there's no discernible difference between the solder jumper and regular resistor footprint after assembly:



Closing and opening the solder bridge manually (with a solder blob) also works pretty well.

Although this is, of course, not comparable to a real production run, I'm now a little more confident in my footprint.

PR #45, opened 2018-05-03
Added Dual Diode (series), SOT-23 package and BAV199

atoav:
Added symbol, unit and entity for Dual Diodes (series) the SOT-23 package and BAV199 part. The package and the part are based on [this datasheet](https://www.infineon.com/dgdl/Infineon-BAV199SERIES-DS-v01_01-en.pdf?fileId=db3a30431400ef6801141cba733104e5)

PR #133, opened 2019-11-26
MMBTA42 300 V SOT-23 NPN transistor

fruchti:
The SOT-23 package is from PR #78.


carrotIndustries:
Bot!

bsilver8192 (approved):
I reviewed this package in preparation to using it.

atoav:
Consider updating SOT-23 from #45 (has a alternate silkscreen as well).

I used (the branch on my fork is called `sot-23`):
```
git checkout sot-23 3d_models/ic/so/SOT-23.step packages/ic/smd/sot/sot-23/package.json
```

carrotIndustries:
That one would also benefit from a sot 23 NPN base part.

fruchti:
> Consider updating SOT-23 from #45 (has a alternate silkscreen as well).

Thanks for the ping! I rebased this PR, pulling in the current SOT-23.

> That one would also benefit from a sot 23 NPN base part.

Is there a safe way to switch to a base part later? I assume recreating the part with the base part referenced, swapping in the old UUID could have some unintended consequences?

RX14:
> I assume recreating the part with the base part referenced, swapping in the old UUID could have some unintended consequences?

It's generally safe to reuse a UUID like that, as long as the entity and package stay the same.

carrotIndustries:
> It's generally safe to reuse a UUID like that, as long as the entity and package stay the same.

I also think that there should be no adverse consequences in the mentioned circumstances. I've also thought about adding something like "insert base part" (any ideas for a  better name?) to the pool manager that does exactly this.

atoav:
> Is there a safe way to switch to a base part later? I assume recreating the part with the base part referenced, swapping in the old UUID could have some unintended consequences?

I did this a hundred times already by hand and never had any issue with it. This is also why am all for @carrotIndustries proposal:

> I've also thought about adding something like "insert base part" (any ideas for a better name?) to the pool manager that does exactly this.

I think something that describes the relationship more clearly would be even better. Maybe something among the lines of:  

- "Inherit from base part"
- "Attach to base part"

and then being able to either select an existing base part, or creating a new one. That would be perfect. I always dealt whit this situation using a text editor for now : )

PR #136, opened 2020-01-18
Add DC/DC Converter and Meanwell xCWN03 series

atoav:
Added the whole Meanwell xWN03 series (3W DC/DC Regulated Converters). See Datasheet [here](https://www.meanwell-web.com/content/files/pdfs/productPdfs/MW/DCWN03/SCWN03,DCWN03-spec.pdf). The 3D model was made with OpenSCAD, rights are no problem there


carrotIndustries:
Bot!

carrotIndustries:
- There's an arduino nano in this PR
 - Having two COM pins in the symbol is confusing. Either have one com pin or COM1/COM2
 - Derived parts don't inherit tags

atoav:
> Having two COM pins in the symbol is confusing. Either have one com pin or COM1/COM2

These are actually two pins for the same COM, but I can name them differently sure

PR #138, opened 2020-01-19
added ATmega328P-AU

endofexclusive:


carrotIndustries:
Bot!

carrotIndustries:
- Power pins should be on top/bottom edges of symbol
 - Package is missing silkscreen outlines apart from pin 1 indication
 - Assy refdes extends past assy outline

PR #139, opened 2020-01-24
Fixing 10118194-0001LF USB Connector

atoav:
Pinout for the 10118194-0001LF was wrong (figured this out with a PCB in
my hand). Either it has been confused with the 10103592-0001LF (same
connector but flipped on it's head) or someone made the mistake of
looking at a male pinout for a female part. Either way, this should fix
it. 

For my PCB I can just get the 10103592-0001LF as a quick fix.

See: 
grafik

### DIFF
--- a/entities/manufacturer/Arduino/Arduino_Nano_v3.json
+++ b/entities/manufacturer/Arduino/Arduino_Nano_v3.json
@@ -1,0 +1,19 @@
+{
+    "gates": {
+        "e0908d75-8c73-47ec-a8b2-ac9fef28c495": {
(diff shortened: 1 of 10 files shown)


carrotIndustries:
Hm, I used this connector in an older revision of https://github.com/carrotIndustries/hubble/ without any issues.

Note that the connector is placed on the bottom side.
2020-01-26-132242_852x482_scrot

atoav:
Wait you're right. The mistake was somewhere else entirely – thanks for the heads up.

PR #140, opened 2020-02-04
Add STM32F030F4P6

fruchti:
Includes a TSSOP-20 footprint as described in the microcontroller's datasheet because the pad sizes differed quite a bit from the existing one. The part is already in use in [a project of mine](https://github.com/fruchti/leshy).

Depends on [the power pin placement issue](https://github.com/horizon-eda/horizon-pool-convention/issues/7) as well as the [package name dash issue](https://github.com/horizon-eda/horizon-pool-convention/issues/9).


fruchti:
Thanks for the merge! Is this also an indirect comment on the referenced pool convention issues?

PR #141, opened 2020-03-03
Add Molex USB Micro B connector 1040170001

frmdstryr:
3d model is from Molex


fruchti:
> 3d model is from Molex

Is it compatible with the pool licence? At first glance it appears to me all of Molex's 3D models are copyrighted and may not be redistributed without their explicit approval.

frmdstryr:
Probably not. 

Maybe horizon should support using URLs for the 3d model so they can be downloaded instead?

PR #142, opened 2020-03-03
Add C&K PTS815 Tactile Switch

frmdstryr:
3d model is from Alex Fedorov on Grabcad https://grabcad.com/alex.fedorov-3

### DIFF
--- a/entities/mechanical/switch/push_switch_spst_off-on.json
+++ b/entities/mechanical/switch/push_switch_spst_off-on.json
@@ -1,0 +1,22 @@
+{
+    "gates": {
+        "1926ff13-c95b-49f9-937c-35ac08ba9efb": {
+            "name": "Main",
+            "suffix": "",
+            "swap_group": 0,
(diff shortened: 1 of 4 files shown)


carrotIndustries:
- There is no 3D model in this PR
- no tags
- omit spaces in MPN for consistency with other parts of this manufacturer

frmdstryr:
It would be nice if it popped up a warning/confirm dialog when a part with that has a 3d model is added to merge but the 3d model isn't included. I didn't realize the merge 3d must be clicked separately.

frmdstryr:
Also might be good to have a rules checker or something before allowing PRs to be pushed.

Is there a way to rename the file? Or will the pool manager handle it if I just rename it (to remove the spaces)?

fruchti:
> Is there a way to rename the file? Or will the pool manager handle it if I just rename it (to remove the spaces)?

You can rename the file and run ‘Update pool’, which will refresh the pool database and update the file name there. Since everything is referred to by UUID, the file name is only cached for faster access.

PR #143, opened 2020-03-03
Add STM32H743 and LQFP-100

frmdstryr:
3d model is from Kicad

### DIFF
--- a/entities/ic/mcu/stm/stm32h743vit6.json
+++ b/entities/ic/mcu/stm/stm32h743vit6.json
@@ -1,0 +1,19 @@
+{
+    "gates": {
+        "5fe65382-4178-4af8-ad33-1ba71375e179": {
+            "name": "",
+            "suffix": "",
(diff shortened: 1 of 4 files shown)


carrotIndustries:
symbol is missing

fruchti:
The unit doesn't include all alternative pin functions but for a part this size I'd personally say this is fine as it would be a lot of work to manually list them all.

There are some issues with the package as well:

* The assembly layer text and pin 1 decoration are the wrong size (should be 1 mm text size, 1.2 mm for the bevel)
* The silkscreen `$RD` text should be near pin 1
* The silkscreen line offset seems to be incorrect (clearance should be 0.2 mm)

You can also feel free to copy the package from PR #76. If you do, it'd be great if you could re-check it and leave a note if there are any issues with it (or if there aren't).

frmdstryr:
Will close and try to use the package from #76.

PR #144, opened 2020-03-06
Added OPA1692 and VSSOP-8 Package

atoav:
This adds OPA1692 in SOIC-8 and VSSOP-8 variants.
3D-model comes from kicad, package messurements come from the footprint in the datasheet: http://www.ti.com/lit/ds/symlink/opa1692.pdf


carrotIndustries:
This PR also adds an Arduino Nano and LME49726MY.

atoav:
The LME one was intentional (as it uses the same package), the arduino was accidental. There must be something I am doing wrong when fetching updates from upstream etc. I will keep an eye out for it in the future.

PR #148, opened 2020-04-12
Adding Barrel Connector Unit/Symbol/Entity, Würth 6941xx301002 package and part

atoav:
This adds a BOSS guitar pedal-style power connector. There are many no-name parts that use the same footprint, but I found a propper one at Würth: https://www.we-online.de/katalog/datasheet/6941xx301002.pdf

The 3D model comes from Sparkfun, see commit message ee76858 for more details on license


carrotIndustries:
This PR also adds some arduino thing. Probably not intended?

atoav:
> This PR also adds some arduino thing. Probably not intended?

This wasn't inteded. I don't know how they ended up there. Deleted them via Github UI (found no way to delete multiple there, so sorry for the many commits)

fruchti:
Looks pretty good! I also like the schematic symbol. I found some issues while reviewing, though:

* Symbol, unit, entity: I'd prefer the name ‘Barrel jack (switched)’ instead of ‘Barrel Connector (switched)’ since the symbol is specifically a female connector. The unit is generic, but you'd be hard pressed to find a switched barrel plug. Although the current convention doesn't specify the capitalisation completely (I should change that), the majority of the pool doesn't seem to be in Title Case.
* Package:
  * The assembly layer text should have a size of 1 mm and be aligned at the assembly polygon edge.
  * There is a weird extra edge in the package layer.
  * The courtyard polygon looks off. If you set the courtyard expansion to 0, it doesn't reflect the outermost package polygon/pad edges.
  * Silkscreen clearance isn't correct (looks like you forgot to add half the line width?).
  * When looking from below, the 3D view shows that clearly either the 3D model's pins or the pads are off:
    2020-04-26-175047_1600x900_scrot
    Also, one pin is somehow black?
* Part:
  * The description should reflect the inner diameter, since it's equally important for barrel jacks. Also, there should be a space before the unit.
  * Have you found a difference between the two orderable MPNs? At first glance, I don't see any but if the two e.g. have a different contact plating, they should be separate parts.
  * I'm not sure if ‘5.5mm’ should be a tag. It's pretty ambiguous.
  * The `th` tag probably was supposed to be `thd`?

I double-checked the pin mapping with the linked data sheet, which is fine. I didn't check the mechanical dimensions and pad positions so far.

You should also probably rebase this PR, since all intermedia commits and their now-deleted parts add to the pool repository size in the end.

atoav:
Okay this adressed nearly all things you mentioned (thanks for the help!). Two things:
- I used the tag `th` nearly everywhere since the very beginning (I think I saw it in one of the existing parts back then), should I still change it?
- I have a rough idea how to rebase things but here I am not really sure, could you give me a pointer?

fruchti:
> * I used the tag `th` nearly everywhere since the very beginning (I think I saw it in one of the existing parts back then), should I still change it?

You're right! I thought about `thd` analogously to `smd` but `th` is much more common in the pool.

> * I have a rough idea how to rebase things but here I am not really sure, could you give me a pointer?

Maybe not the most elegant way to do it, but with

```shell
$ git rebase -i master
```

you can bring up an editor which should have `pick` for each commit. Replacing this action with an `f` (for fixup) ‘merges’ the commit with the previous one and files added and deleted within the same commit shouldn't be in the repository. I don't know enough about git to know how to remove files from previous commits directly, though.

atoav:
Bot!

PR #152, opened 2020-05-20
Add SOT-23-6 Package

frmdstryr:
Duplicated from the SOT-23-5 and added the 6th pin. 3d Model is from KiCad


frmdstryr:
horizon-sot-23-6-schm
horizon-sot-23-5

PR #154, opened 2020-05-22
Add VSON-8 3x3mm Package

frmdstryr:
Model is from KiCAD


carrotIndustries:
- exposed pad should be named "PAD"
- assembly and package outline are swapped
- assembly refdes must be inside of of assembly outline

PR #155, opened 2020-05-22
Add 2x5 and 2x7 1.27mm pitch generic vertical pin headers

frmdstryr:
3d models are from KiCAD


frmdstryr:
Uhh... the rules check threw an error with the 2x7, after looking again the spacing seems to be messed up with this. 

I used a 1.27mm spacing in the footprint generator but they are actually alternating from 1.3mm to 1.25mm, not sure what is going on here?

PR #157, opened 2020-06-11
Add SOT-23-3 based on existing SOT-23-5

nzottmann:
SOT-23-3 Package and 3D Model, created by removing two pads/pins in the existing SOT-23-5 Package and updating the pin numbers


carrotIndustries:
Bot!

carrotIndustries:
It's the wrong way round compared to the standard IPC definition.

PR #159, opened 2020-06-13
Add MCU STM32F030CCT

jue89:
* (L)QFP-48 package
* symbol
* part


carrotIndustries:
Bot!

carrotIndustries:
Bot!

carrotIndustries:
- Assy refdes text extends past assy outline
 - RST pin should go near the top edge

jue89:
Thanks for the review. For clarification - did I get you right?

rst
refdes

carrotIndustries:
> Thanks for the review. For clarification - did I get you right?

2× yes

jue89:
Bot!

jue89:
Bot!

carrotIndustries:
Since this thing comes in multiple variants in the LQFP48 package, it makes sense to have a common base part and have the actual parts inherit from the base part.

jue89:
Yes I know. I planned to do so in the first place. But If you look deeper in the [datasheet](https://www.st.com/resource/en/datasheet/stm32f030f4.pdf), you notice that the STM32F030CC has additional power supply pins. Its pinout doesn't match to the other MCUs of this family.

Screenshot from 2020-09-20 13-25-29

carrotIndustries:
Bot!

carrotIndustries:
>    x Top Package near Top Silkscreen

Move the silkscreen farther away from the package body.

jue89:
Bot!

jue89:
Bot!

jue89:
Bot!

jue89:
And finally the bot is happy :)

jue89:
@carrotIndustries is there something missing or you're worrying about?

jue89:
The bot is still happy! But I think I'll move the BOOT0 to left side as well ... it's functionality is strictly coupled to NRST.

jue89:
Bot!

jue89:
See the part in action:
Screenshot from 2021-04-07 15-48-42

RX14:
The recommendation for symbols inverted signals is that they have either the "N" (or `~`) prefix or the inverted circle on the pin in the symbol, not both.

jue89:
I haven't picked that name. It was selected by STM. I've added the circle to make it obvious that NRST is low active.

fruchti:
I’m not sure if ‘GPIO’ should be included as an alternate pin function? I mean, it could be useful as a marker in the schematic, but I haven’t seen it anywhere so far.

A few small things currently not agreeing with the convention:

- Prefer perpendicular power pin name orientation in the symbol if possible (so VDD, VSS, etc. are in legible orientation instead of rotated).
- The package should be called ‘QFP48’ (without the dash).
- The package file name should keep the capitalisation.
- At first glance, the package’s silkscreen outline seems to have more clearance than 0.2 mm. Did you use the silkscreen generator, because it should handle this package just fine.
- The silkscreen refdes should be near pin 1. Looking through the pool just now, none of the existing QFPs does this, so maybe this is a point for discussion (otherwise, I’ll take it up to fix it for the existing packages).
- The assembly layer refdes should be anchored across from pin 1.

jue89:
Bot!

jue89:
Thank you for your in-depth review. Everything, you've mentioned should have been addressed.

The GPIO marker came from the STM XML file describing this MCU. I removed all of them. The other STM32 MCUs in the pool don't have this, as well.

fruchti:
Perfect, thanks! I had a quick look at the data sheet and checked the most important pins, which match; so nothing seemed to have gone wrong there.

Two even more minor things I noticed:

- The value for the part could be ‘STM32F030CCT’ without the 6, as there is only one temperature grade anyway.
- The part description could say ‘with 256 KB flash’ explicitly, since the specific part has that amount.

But as these are just nitpicks, I’d declare this ready to merge (feel free to add the label if you don’t want to add another commit just for this).

jue89:
I addressed both nitpicks. I clarified the information regarding the flash size (*up to* doesn't make sense for just on part ;)). The flash size is my reason to select this MCU - so I didn't removed it.

RX14:
This isn't something you have to do, but I like the terse description style used for passives: "ARM-based 32-bit MCU, 256KiB flash"

carrotIndustries:
As far as I can tell the C6 and C8 variants share the same package and pinout, so a part structure like this makes sense:

 - STM32F030Cx
   - STM32F030C6
   - STM32F030C8
   - STM32F030CC

Any reason not to do this?

jue89:
> Any reason not to do this?

Unfortunately, the STM32F030CC is a bit of an oddball: https://github.com/horizon-eda/horizon-pool/pull/159#issuecomment-695776185

> This isn't something you have to do, but I like the terse description style used for passives: "ARM-based 32-bit MCU, 256KiB flash"

Sure. Good idea. I added information I find important when selecting an MCU. In the first place I was lazy and just copied the datasheet's headline 🤪 (Like the other did - maybe we'll should to some housekeeping there. Wo don't want to be their marketing department, don't we?)

jue89:
FYI: I produced 3 different PCBs with this part. I think pin-pad-mapping should be fine.

RX14:
looks perfect to me now

Generating coherent part/base part hierarchies for large, complex series of MCUs is always going to be an uphill battle imo, and adding base parts later *is* possible without changing UUIDs, so this is great to merge now.

jue89:
> Generating coherent part/base part hierarchies for large, complex series of MCUs is always going to be an uphill battle imo, and adding base parts later _is_ possible without changing UUIDs, so I'm onboard with merging this now.

Do you see any possibility to group this part with other STM32? Afaik there is no other part with the same pinout and package. There isn't even an alternative with a different temperature range.

RX14:
> Do you see any possibility to group this part with other STM32?

No, what you have here is fine. Let's merge this.

jue89:
Thank you all 🎉

PR #160, opened 2020-06-13
Add CAN Transceiver MCP2562

jue89:
* symbol
* part

### DIFF
--- a/parts/ic/transceiver/can/MCP2562-ESN.json
+++ b/parts/ic/transceiver/can/MCP2562-ESN.json
@@ -1,0 +1,74 @@
+{
+    "MPN": [
+        false,
+        "MCP2562-E/SN"
+    ],
+    "datasheet": [
+        false,
+        "http://ww1.microchip.com/downloads/en/DeviceDoc/20005167C.pdf"
+    ],
+    "description": [
+        false,
+        "High-Speed CAN Transceiver"
+    ],
+    "entity": "c9664352-6856-4fb8-94f0-6390ac36d339",
+    "inherit_model": false,
+    "inherit_tags": false,
+    "manufacturer": [
+        false,
+        "Microchip"
+    ],
+    "model": "00000000-0000-0000-0000-000000000000",
+    "orderable_MPNs": {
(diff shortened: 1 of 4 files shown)


carrotIndustries:
Bot!

jue89:
Bot!

bsilver8192:
I looked through this to use in a project of mine. LGTM, I'll update once I get boards to try it out on.

parts/ic/transceiver/can/MCP2562-ESN.json
```diff
@@ -18,11 +18,10 @@
         false,
         "Microchip"
     ],
-    "model": "00000000-0000-0000-0000-000000000000",
+    "model": "b99785a3-c0da-4658-969d-6287cec3be05",
     "orderable_MPNs": {
-        "340cc66d-a0db-4e3a-8c2a-3fa498baf856": "MCP2562-H/SN",
         "ab7348d5-75d4-46f1-ba3a-e1b1313b1479": "MCP2562T-E/SN",
-        "abcc3ee9-a73c-4e51-8b4c-3547651109f7": "MCP2562-H/SN"
```
bsilver8192: I didn't notice the duplicate before, but I think `MCP2562-H/SN` is a valid orderable number number (high-temperature).

jue89:
Thanks for testing this part out! I've produced a few PCBs with this part in place - works as expected :)

I've polished part info:
* I removed the other temperature range - they are other parts
* I put the most important info into the description instead of the datasheet's title (which has been influenced by Microchips' marketing, I guess?)

parts/ic/transceiver/can/MCP2562-ESN.json
jue89: Yep, you are pretty fast with your reviews ;) If I understood the orderable MPNs correctly, they should just differ in packaging. The temperature range is a property of the device it-self. Or am I wrong?

bsilver8192: Hmmm, I think you're right. I looked at the convention, and it even uses temperate range as an example:
> Use inheritance for only slightly differing parts (like different packaging, temperature range, etc.).

carrotIndustries:
Bot!

PR #171, opened 2020-09-29
Add small resistor and capacitor symbols

RX14:
These symbols are similar to kicad's R_small and C_small symbols, for
use in tight schematics.

Here's an example of their use (R2-4, C7):
poomf-linux-2020-09-29T14 24 38


RX14:
Bot!

RX14:
Well that didnt help much

carrotIndustries:
The thing is that having these will always bring up the select symbol dialog when placing a resistor which is really annoying. Before merging these, the map symbol tool helper should remember symbol choices.

RX14:
> The thing is that having these will always bring up the select symbol dialog when placing a resistor which is really annoying. Before merging these, the map symbol tool helper should remember symbol choices.

Yeah, that's something I noticed but forgot to mention in the PR head. I like that solution!

carrotIndustries:
> Before merging these, the map symbol tool helper should remember symbol choices.

https://github.com/horizon-eda/horizon/commit/30975bd3055c2b986b14ab8fa426aa03057f1537  adds this

Having a look at the symbols, how about also making the text smaller? The small symbols look somewhat out of proportion next to the regular-sized text.

RX14:
I've pushed smaller capacitor and resistor text sizes. I'm not 100% on the change but I think it is better, especially in situations like this

carrotIndustries:
I'm having second thoughts about whether we really want to have small symbols in the default pool as these might lure people into making messy schematics. Once we have proper support for using multiple pools in projects however, I see nothing wrong with having a "small symbol" pool that people can use if they _really_ want to.

RX14:
I think people will make messy schematics whether you try and prevent them or not, and they'll probably just make messy symbols of their own if you don't make them easy to get. Kicad has these symbols, and they're popular, so I think people (like me...) will just make their own if they're not in the upstream pool.

carrotIndustries:
Now that we have support for multiple pools, I suggest you put them in their own pool that includes the main pool. When done, add it to https://github.com/horizon-eda/horizon-pool-index

RX14:
Yeah, I think I'm having a change of heart on this too, I think my new designs will use larger symbols and include more information on the schematic instead of trying to be more compact

PR #175, opened 2020-11-25
Generate Murata GRM capacitors, with script

RX14:
This was pretty fun


RX14:
Any chance to get this merged?

carrotIndustries:
The part numbers appear to be missing the packaging suffix and thus can't be looked up by the partinfo integration. These should go into the orderable MPN list.

Also some MPNs such as GRM188R71C104KA01 appear to be at the tail end of their lifecyle (to be discontinued/obsolete). We don't want such parts cluttering the pool.

RX14:
A lot of "not recommended for new designs" parts continue to be highly stocked by distributors. I'd rather generate everything when it's possible to.

I'll have a look at fixing the orderable MPN list though.

RX14:
Right... I forgot that the CSV provided uses "#" for all the packaging suffixes. To get the valid packaging for each part you need to perform a http request per part, which isn't feasible for 10k parts

carrotIndustries:
https://github.com/horizon-eda/horizon/commit/ac48399234f0073d8701275ff488a86dc326f00b makes the missing suffixes somewhat acceptable as the partinfo API now returns results even if the suffix is missing

RX14:
So is this ready to merge? I think something is better than nothing?

PR #180, opened 2021-01-16
ATMega32U4 and QFN-44

skuhlabs:
Adding the ATMega32U4 MCU along with the QFN-44 footprint and 3D Model.  Datasheet is located: http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-7766-8-bit-AVR-ATmega16U4-32U4_Datasheet.pdf, and further documentation refer to JEDEC drawing MO-220 variation VKKD-1.


frmdstryr:
Bot!

carrotIndustries:
Bot!

carrotIndustries:
Seems like you forgot to add everything but the 3d model.

starlord-a:
> Seems like you forgot to add everything but the 3d model.

Strange, I thought had everything select - I will do it again in a bit. Thanks!

starlord-a:
Apparently when I did this before I forgot to add nearly everything, not sure how I managed that, but here we go again.  Also, I placed the REFDES and VALUE in their proper spots.

carrotIndustries:
Bot!

carrotIndustries:
Bot! learned some new tricks, let's see how these work...

starlord-a:
Will correct the above issues soon (tm)

fruchti:
I have only checked the bot’s output and haven’t compared anything to the data sheet so far. Apart from what the bot said:

Package:

* The assembly reference designator has to fit into the assembly polygon. Scale it down in 0.1 mm increments (decrements?) until it fits.
* The assembly polygon should have a 1.2 mm large bevelled edge at pin 1.
* The $RD silkscreen text should be near pin 1.
* The silkscreen pin 1 marker seems unusual? It’s certainly different to other QFNs in the pool, but that’s not necessarily a bad thing.
* The convention currently calls for omitting the dash in the package name (‘QFN44’). There’s [an open issue](https://github.com/horizon-eda/horizon-pool-convention/issues/9) about this because it might, IMO, not be the most intuitive rule.

Unit:

* Is RESET really bidirectional? I don’t see any alternate functions listed.
* Assuming UCAP is used for an external capacitor to stabilise an internal regulator, I’d label it either power output or passive.

Symbol:

* Make sure it is expandable.
* I have no experience with this particular part, but maybe reordering the pins could benefit ergonomics? I’m thinking of placing VBUS near D+/D- or the XTAL pins near the bottom.
* For better legibility, I’d personally prefer the ‘perpendicular’ pin name orientation for pins on the top and bottom. I (very) recently added this to the convention but as this has been mostly decided by me so far I’m happy to discuss if you disagree.

PR #182, opened 2021-01-16
Add L293 Quad Half-H Drivers

atoav:
This adds the TI Quad Half-H Drivers [L293 and L293D](https://www.ti.com/product/L293)

### DIFF
--- a/entities/ic/driver/L293.json
+++ b/entities/ic/driver/L293.json
@@ -1,0 +1,21 @@
+{
+    "gates": {
+        "273fc03e-4af9-404a-9498-4959b526ed7e": {
+            "name": "Main",
+            "suffix": "",
+            "swap_group": 0,
+            "unit": "4d1c6a24-d000-449f-a09f-a64602aef80d"
+        }
+    },
+    "manufacturer": "Texas Instruments",
+    "name": "L293",
+    "prefix": "U",
+    "tags": [
+        "driver",
+        "half-h",
+        "ic",
+        "quad"
+    ],
+    "type": "entity",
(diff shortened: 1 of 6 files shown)


atoav:
bot!

carrotIndustries:
Bot!

> bot!

The bot is case sensitive.

atoav:
Ahh I will fix the datasheet link. Google changed that copy behaviour to MITM themselves into everything it seems

carrotIndustries:
Bot!

carrotIndustries:
LGTM overall
 - The L293 is also available from ST. by making the unit/entity generic, we could reuse them for the one from ST
 - The symbol is a bit confusing, inputs on the left and out on the right would be more clear
 - Not sure about the added value of specifying text placements. If people end up placing that symbol horizontally, they're most likely doing the wrong thing.

atoav:
> The L293 is also available from ST. by making the unit/entity generic, we could reuse them for the one from ST

Sounds like a good idea

> The symbol is a bit confusing, inputs on the left and out on the right would be more clear

I just followed the datasheet, but sure that could easily be re-arranged

> Not sure about the added value of specifying text placements. If people end up placing that symbol horizontally, they're most likely doing the wrong thing.

Ah, I thought I _had_ to do that. Should I remove it?

carrotIndustries:
> Ah, I thought I had to do that. Should I remove it?

From my perspective, they're only really required if the symbol is expected to be placed in all orientation such as a resistor or a transistor. Since yours all turned out correctly, we can keep them.

fruchti:
Any particular reason for the pin spacing in the symbol? Using 2.5 mm like other box-type symbols do would make it quite a bit more compact.

PR #184, opened 2021-01-16
Add Highly PB6149L LED illuminated push switches

atoav:
This adds some of the Highly Electric [PB614 series](https://www.highly-elec.com/PB614-series.html) push switches.
I only added switches with actuator shape 9, without frame, standard type and without icons and such.

Switch manufacturers are always a little inprecise with their data sheets but I crosschecked with the switches I have on my desk.

Didn't create a 3D model tho, but maybe I will do that soon.


carrotIndustries:
Bot!

carrotIndustries:
Multi-line values overlap

PR #187, opened 2021-01-17
M3 Mounting Hole

skuhlabs:
Nothing special, just a mounting hole sized for M3 fasteners.


starlord-a:
Some weird things were happening when I was trying to update the pool so I recreated this part and resubmitted it.

PR #189, opened 2021-01-21
Add STM32F405RGTx MCU

skuhlabs:
Adding the part mentioned in the title, datasheet is linked in the part. Using standard QFP-64 package, not added.

### DIFF
--- a/entities/ic/mcu/stm/STM32F405RGT6.json
+++ b/entities/ic/mcu/stm/STM32F405RGT6.json
@@ -1,0 +1,21 @@
+{
+    "gates": {
+        "3845ddad-5110-4e0d-b6df-0dc3ca833ead": {
+            "name": "Main",
+            "suffix": "",
+            "swap_group": 0,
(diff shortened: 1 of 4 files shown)


carrotIndustries:
Bot!

carrotIndustries:
- Description is too long
 - Pins are missing alt names, there's a script in the horizon repo that extracts them from the STM32CubeMX application
 - We already have other STM32 MCUs in the pool, @fruchti  do you think it makes sense to have conventions specific to particular popular MCU families?

fruchti:
> * We already have other STM32 MCUs in the pool, @fruchti  do you think it makes sense to have conventions specific to particular popular MCU families?

IMHO this could quickly get overly verbose. Having rules like ‘NRST must be on the top left for STM32 symbols’ would make the symbols more uniform but would be too specific for a general convention for my taste.

I’d much rather add a general rule to check existing parts/PRs of the same series and how they approach the symbols (how pins are grouped and positioned within the symbol, whether it is split up into multiple gates, naming conventions).

PR #192, opened 2021-01-27
add 93LC46B EEPROM

jue89:
* part
* symbol

### DIFF
--- a/entities/ic/manufacturer/microchip/93xxx6AB.json
+++ b/entities/ic/manufacturer/microchip/93xxx6AB.json
@@ -1,0 +1,19 @@
+{
+    "gates": {
+        "366ee996-504f-441a-9a24-9c399a9d3671": {
+            "name": "Main",
+            "suffix": "",
+            "swap_group": 0,
+            "unit": "db1607d2-9f77-4685-a34c-ac2ec6aacaf1"
+        }
+    },
+    "manufacturer": "Microchip",
+    "name": "93xxx6A/B",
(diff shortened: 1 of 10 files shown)


jue89:
Bot!

carrotIndustries:
As far as I can tell, this device is part of a greater family, so it probably makes sense to introduce a base part and inherit from that one.

RX14:
The symbol perhaps looks even more generic than just one manufacturer or logic family as well. Though, that could easily be renamed later

jue89:
If the symbol should become a common serial flash I would rename the pins: VCC and VSS are the names from the datasheet. I would expect the combination VDD/VSS or VCC/VEE ... in a common symbol I'd select VCC/GND.

jue89:
Symbol ist generic. I generated a base part for 1kBit EEPROMs.

I added a base part for the EEPROMs with larger memory in the first place - but that was a little bit dumb - the datasheet is wrong for parts with more memory ... or is this the right solution?

- Basepart 93xxx6AB/SN (pin mapping, entity)
   - Basepart 93xx46AB/SN (datasheet link)
      - Part 93LC46B/SN

jue89:
Bot!

RX14:
LGTM

carrotIndustries:
@RX14 How about renaming the `CS` pin to `~CS` as it's low active?

RX14:
<abbr title="Looks Even Better To Me Now">LEBTMN</abbr>

PR #194, opened 2021-02-18
Add relay G5Q-1A-EU DC12

jue89:
- SPST-NO symbol, unit, entity
- package
- part

- [x] Fix datasheet link


fk0815:
Bot!

carrotIndustries:
Bot! (the bot now dislikes reichelt datasheet links)

carrotIndustries:
Bot!

jue89:
Bot!

jue89:
Since this PR has been created during a live stream in almost no time, it was far from perfect. Some aftermath:

- Part: The datasheet link has been fixed
- Symbol: Some polishing ... the dashed line now looks better at least to me.
- Unit: Changed the pin names of the switch to `COM` an `NO`. This should be much more descriptive than bare numbers.
- Assembly: Removed the pin 1 decorator of the outline and added the characteristic pins. (Are polygons the right choice for that?)

Furthermore, I think it would make sense to create a G5Q-1A base part and inherit its specific versions (G5Q-1A-EU DC12, ...). Is this the *right* way?

carrotIndustries:
> I think it would make sense to create a G5Q-1A base part and inherit its specific version

Yes, that's the proper way.

jue89:
Bot!

fruchti:
General question (also concerns #197): What should the relay symbol look like? I’ve never seen it with a box around it, at most with a double dashed line connecting coil and contacts. Personally, I’d prefer the coil and contacts to be separate units because they usually end up in quite different areas of the schematic. IMHO, it’s also not unusual to have them separate so this won’t be surprising or confusing to ‘outsiders’.

I have a relay unit/symbol set ready in a project pool I can add as a new PR if there’s interest for a symbol in this style.

jue89:
I aligned the symbol with the already-existing solid state relay. I got the single-dashed line from the IEC symbol - but to be honest the coil isn't IEC-ish. So maybe I should remove diagonal line from the coil.

And personally, I prefer to have the relay in a single unit. I draw a net line or bus from signal source to the relay instead of having contacts and coil in different regions of the schematic. When reading the schematic you want to have a obvious connection between the contact, the coil and the driver of the coil.

fruchti:
Alright, I would’ve preferred pins in line with the coil/switch symbols but as the majority agrees I’m also fine with this style.

PR #195, opened 2021-02-20
Make all generic connectors expandable

jue89:
The schematic editor allows to use custom names for pins. I'd like to use this feature - but I cannot expand the symbol.

This PR fixes this. The changes have been made using some sed magic:
`sed -i 's/"type": "symbol",/"type": "symbol", "can_expand": true,/' *`

(I don't know if its a good idea to trigger the bot :D)


jue89:
Thank you!

PR #196, opened 2021-02-22
Add USB-C Receptacle for USB2.0

jue89:
Added package and part. Made existing symbol, unit and entity generic.


jue89:
Bot!

jue89:
Thanks for reviewing :)

The receptacle already has been soldered on a PCB which used this part/footprint. It works like a charm.

PR #197, opened 2021-02-25
Add Meder Reed relay SIL05-1A72-71D

dl8dtl:
Meder Reed relay in SIL case, 5 V coil voltage, integrated free-wheeling diode


jue89:
Bot!

carrotIndustries:
# Symbol 
- Refdes & value should move to the very left
- Relay coil box should be vertically centered

# Package
- Square pin 1 isn't standard among other packages in the pool
- Where's the 3D model from? What's the relation to #198?
- Silkscreen is missing pin 1 indicator
- Name could be more generic

# Part
This part is most likely a member of a larger family, so it makes sense to add a common base part and have this part inherit from the base part.

dl8dtl:
Update:

* make symbol symmetric
* align $REFDES and $VALUE left
* make pin 1 round
* add pin 1 marking in silkscreen
* find family datasheet, make package name more generic

dl8dtl:
Well, I missed to add the 3D model to the initial pull request. So yes, #197 and #198 belong together.
Found a family datsheet now.

RX14:
Now #194 is merged, I think the symbol could be tweaked slightly to exactly match: the stripe on the coil and the single-dashed line.

RX14:
Bot!

RX14:
Also, the package should have pin 1 marked using something other than the line. The arrow symbol from headers looks like a good bet.

PR #198, opened 2021-02-25
Meder SIL05-1A72-71D 3D model

dl8dtl:
Created after drawing in datasheet.


RX14:
This really should be merged into #197. Horizon has built-in PR editing tools now!

PR #199, opened 2021-02-28
Add Kailh PG1350 switch

tadfisher:
Kailh "Choc" PG1350 switch. Footprint is reversible for split keyboards, supports hot-swap pin receptacles, and is compatible with Cherry MX switches.


RX14:
Bot!

RX14:
Are these footprints look pretty specialized beyond supporting the PG1350?

tadfisher:
@RX14 They support Cherry MX hotswap sockets as well. I can make these compatible with PG1350 only if that makes the review process simpler.

RX14:
I think it'd be good to have the most basic package as the default, and then have a more complex package which supports more than the part it's attached to as an alternative, since horizon supports alternative packages.

PR #200, opened 2021-03-12
Add narrow variant to PJ301M-12 package

atoav:
This is just an additional package variant for the PJ301M-12 which we already have in the pool (and which I used in many designs with some success). Pin 1 is closer to the body in that variant, which is totally feasible because the leg for pin 1 is roughly 11 mm long and flexible enough:  
2021-03-12 09_39_21-Window  

This can be helpful in some tight spaces (e.g. like the one I am having now : )


carrotIndustries:
Bot!

PR #201, opened 2021-03-14
Add ideal diode MAX40200

jue89:
- symbol/unit/entity for ideal diodes with enable pin
- part MAX40200AUK+

### DIFF
--- a/entities/ic/switch/ideal-diode-en.json
+++ b/entities/ic/switch/ideal-diode-en.json
@@ -1,0 +1,20 @@
+{
+    "gates": {
+        "61f4dc85-b963-41ee-9f3e-89e50c03c40f": {
+            "name": "Main",
+            "suffix": "",
+            "swap_group": 0,
+            "unit": "996b6d7c-9c3e-44e6-bcf0-fcefecc8e348"
+        }
+    },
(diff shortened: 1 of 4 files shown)


jue89:
Bot!

carrotIndustries:
Wouldn't it make more sense to put the input left and output right?

jue89:
> Wouldn't it make more sense to put the input left and output right?

Yes, I fully agree. I was biased by the schematic I needed the symbol for :D I'll mirror the symbol.

jue89:
Bot!

RX14:
For this part, it might be ideal to support every orientation, or at least left/right, by fixing the text locations to make that look nice.

jue89:
Bot!

RX14:
The rest are a bit funky, but normal and mirrored are fine so this looks OK to me

jue89:
> I wouldn't complain if the 180degree ones were improved though.

Yeah - but in that case the GND ping is in the top section of the symbol. That's something that shouldn't be done in my opinion - no one expects GND to be on top when reading schematics ;)

RX14:
ah, you're right

it might be neat to have more orientation-dependent symbols but thats for another day

PR #202, opened 2021-03-17
Convert passive packages to rounded rectangle pads

RX14:
IPC-7351C recommends rounded rectangle pads, and kicad's default passives are now roundrect too. The corner radius is `min(0.25, min(width, height)/4)`.


rroohhh:
Bot!

PR #185, opened 2021-01-17
Molex 1054500101 USB Type-C Receptable

skuhlabs:
Molex Type-C Receptable (1054500101), Manufacturer Page: https://www.molex.com/molex/products/part-detail/io_connectors/1054500101, Datasheet: https://www.molex.com/pdm_docs/sd/1054500101_sd.pdf.

PR #203, opened 2021-03-30
Add Molex 105400101 USB Type-C connector

RX14:
Tested this on a board of my own, works great.

Closes #185.

### DIFF
--- a/entities/connector/usb/type-c-shield.json
+++ b/entities/connector/usb/type-c-shield.json
@@ -1,0 +1,20 @@
+{
+    "gates": {
+        "68d7944b-de7b-4e29-a7ec-85044be94ff1": {
+            "name": "Main",
+            "suffix": "",
+            "swap_group": 0,
+            "unit": "3ce158b1-63f0-4bfa-8c58-ae811485125d"
+        }
+    },
+    "manufacturer": "",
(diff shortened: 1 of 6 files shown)


carrotIndustries:
Bot!

RX14:
The "top copper near top silkscreen" clearance here seems to be a bug in horizon to me. Or at least, the arrow pointing to where the intersection is seems erroneous

carrotIndustries:
How about setting vbus, gnd and shield pins to perpendicular?

RX14:
Bot!

jue89:
Duplicate / alternative of #185

RX14:
Ah, might be worth taking the model from there.

RX14:
I'm happy to take the model from #185 if others agree. The model doesn't *look* to be from molex's website so can probably be included.

RX14:
Also: Molex has changed their part naming to exclude the dash, but most stocking distributors still have the dash. Other parts in the pool have the dash, this one has no dash, should I add the dash to this part? I think all molex parts going forward should be one or the other.

RX14:
The model in #185 has "License CC BY-ND 4.0" in it's header. That's incompatible with this repo, so I can't include that model.

PR #205, opened 2021-04-01
Add Molex 73251-1150 SMA connector

RX14:
Tested on a recent board


RX14:
Bot!

RX14:
The only question I have about this part is the vias being included in the package. They're in the recommended footprint, but it might be worth at least having a variant without them?

carrotIndustries:
> They're in the recommended footprint, but it might be worth at least having a variant without them?

I think so, so users can add vias as they see fit. Also, vias baked into the package might be in the way of what's on the board.

RX14:
Bot!

I've added a variant without vias. It's the default right now, but it could be switched.

PR #207, opened 2021-04-01
Add 2520 SMD crystal package and 32MHz crystal

RX14:
2520 is quite a common MHz crystal SMD package.


carrotIndustries:
Bot!

carrotIndustries:
Bot! again. Triggering the bot in fast succession doesn't work that well...

RX14:
I checked 2520 packages between multiple manufacturers btw, and this recommended land pattern seems to be close enough for all of them to work.

PR #210, opened 2021-04-05
Add XR21V1410 USB UART

RX14:
Nice cheap widely available USB UART chip using CDC device class. Tested on a board and works great!

### DIFF
--- a/entities/ic/interface/usb/maxlinear/xr21v1410il16-f.json
+++ b/entities/ic/interface/usb/maxlinear/xr21v1410il16-f.json
@@ -1,0 +1,19 @@
+{
+    "gates": {
+        "91b3e82d-f86a-42da-8334-96fc28cbb8e3": {
+            "name": "Main",
+            "suffix": "",
+            "swap_group": 0,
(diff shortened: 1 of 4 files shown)


carrotIndustries:
Bot!

jue89:
I'd put RX and TX on the right side of the symbol. They are strongly coupled to the GPIOs, since they can be used as handshake signals. Furthermore, my inner mock is slightly triggered that the GND pin got its own *row* and the VCC pin shares half of its *row* with the RX pin ;)

RX14:
Yeah, I think those are good points. The symbol is a bit messy, it should probably be expanded a fair bit more.

RX14:
Bot!

Much happier with this version

carrotIndustries:
LGTM, though description could use a bit less ALLCAPS. Maybe flag words with more than 5 letters and all of them being uppercase as a warning?

RX14:
I must have been tired when I made that part - I don't usually copy directly from the datasheet!

RX14:
Done! The diff is simple enough it's not worth calling the bot.

PR #211, opened 2021-04-05
Touch up QFN-16 package

RX14:
The QFN-16 package in the library had some issues appear when I had a board manufactured using it.

Most importantly, the soldermask didn't correctly flow down the sides of the pads. Also the silkscreen was off.

I've regenerated this package using the IPC generator, which has the correct pin sizing, allowing soldermask to flow between the pins. I also regenerated the silkscreen to be 0.2mm from the part outline as guidelines recommend.


RX14:
Bot!

PR #212, opened 2021-04-06
Add ATmega16/32

fk0815:


fk0815:
Bot!

RX14:
It's more typical to have a generic "ATMega16/32" part and then have the specific "ATmega16A-AU" and "ATmega32-16AU" as children of that part. That way the atmega32 isnt arbitrarily chosen to be the child of the atmega16.

I would also suggest looking into [having a unit for each port](https://docs.horizon-eda.org/en/latest/pool-elements.html#units), and a power unit, so you can move each port around the schematic seperately. Otherwise the symbol gets unwieldy like the one above.

fk0815:
Bot!

RX14:
- The generic part is specific to the TQFP package, so I think it should be called something similar to `ATmega16/32-AU`
- I'm unsure about the unit naming. Are these units used beyond the ATmega16/32 parts? Otherwise I think the unit names should have ATmega16/32 in.
- From reading the datasheet, alternate names like `XCK/T0` appear to be two seperate alternate functions, so they should be listed as two seperate alternate pin names (replace the `/` with a space).
- Since `~` now renders an overbar, and SPI `SS` is listed in the datasheet with an overbar, it might be well to call that pin `~SS`. `~RESET` can be treated in the same way. This is up to you.
- The power symbol is a bit messy. The power pins at the top should be centred. It might also look nicer if you widened it out a bit, and placed the pins on each side opposite from each other. Experiment and see what looks nicest to you.


Almost there!

fk0815:
Bot!

fk0815:
> * I'm unsure about the unit naming. Are these units used beyond the ATmega16/32 parts? Otherwise I think the unit names should have ATmega16/32 in.

I reworked the PR to add your comments. The idea behind the naming of the units is that they can be re-used without a naming conflict for other AVR like devices that have the same ports with the same alternate functions layout. Therefore I would like to keep them as they are.

RX14:
LGTM

RX14:
Ah! Few things I missed from the [symbol guidelines](https://github.com/horizon-eda/horizon-pool-convention):
- the power symbol needs horizontal power pin texts
- The port symbols should enable expansion, and be less wide.
- The port symbols should be symmetric about the origin

fk0815:
Bot!

RX14:
This looks all good now, could make the power symbol a bit less tall but I won't hold up merging for it.

carrotIndustries:
I'm not quite sure how I should feel about a MCU with not that many pins been torn up into 5 units. I've never seen it done this way in any schematic. My personal preference would be to keep it all in one unit.

RX14:
Ah, my bad... It was my preference to do it this way but I'll remember this for future entities.

Apologies to @fk0815, I'm happy to finish this off if you'd prefer.

fk0815:
Bot!

carrotIndustries:
Thanks for fixing, one last thing: the XTAL pin's aren't passive. According to the datasheet, XTAL1 is input, XTAL2 is output

fk0815:
Bot!

PR #213, opened 2021-04-06
Added MCP1640 device

fk0815:
Signed-off-by: Frank Kunz <mailinglists@kunz-im-inter.net>

### DIFF
--- a/entities/ic/regulator/microchip/mcp1640.json
+++ b/entities/ic/regulator/microchip/mcp1640.json
@@ -1,0 +1,19 @@
+{
+    "gates": {
+        "8e3a69c1-ca74-4bf3-b577-9121c7d0fdd3": {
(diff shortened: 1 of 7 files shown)


fk0815:
Bot!

fk0815:
There is just a warning about the symbol, but no details what is the problem.

fk0815:
Bot!

RX14:
My only suggestion is to add the "regulator" tag as well as the others

carrotIndustries:
- no need for orientation-dependent text placements
 - pins on top/bottom should use perpendicular text
 - en and vfb pins should be at the the same height

RX14:
It feels like the whole symbol could be a bit less tall as well

fk0815:
Bot!

PR #214, opened 2021-04-06
Add RFM9x device

fk0815:


fk0815:
Bot!

fk0815:
Bot!

fk0815:
Bot!

RX14:
Ditto with the suggestion on the derived parts tree: there should be a generic RFM9x part, which is the parents of the other parts.

I'd also recommend moving the DIO pins to the right hand side of the symbol.

carrotIndustries:
- Making the symbol bit taller, 3v3 and GND could become vertical
 - Orientation-dependent texts aren't needed for ICs
 - Datasheet is 404

fk0815:
Bot!

RX14:
Everything's great here except the symbol. `GND` and `3V3` should be in horizontal orientation, and the height needs to be collapsed so that there's less wasted space along each side.

Also, from the [pool conventions](https://github.com/horizon-eda/horizon-pool-convention):
> Do not use a ‘inverted’ decoration for pins whose name already indicates inversion (n or / in front, overbar, etc.)

fk0815:
Bot!

RX14:
I'm unsure of @carrotIndustries's exact thoughts on this, but i'm aware he might prefer `~RESET` here instead of the symbol decoration. We should probably make the guidelines clearer whatever the answer is.

RX14:
Other than pending questions on symbol pin decorations, this looks good to me.

carrotIndustries:
- The pin's called `RESET` in the datasheet, so I think adding the dot is fine in that case
 - On the `NSS` pin however, it's redundant
 - Pads should be numbered 1...16, easy to fix using the renumber pads tool.

RX14:
Perhaps the `NSS` pin be renamed `~SS` then?

RX14:
@carrotIndustries and I have had a chat on IRC and came up with this:

- all signals starting with `/`, `~` or `N` in the datasheet should be standardized to `~`
- pin decorations such as the arrow for clock or the circle for inverted should *only* be used on logic symbols, not ICs
- It's up to the author whether to add the `~` prefix to signals which are inverted, but are not shown prefixed in the datasheet

fk0815:
Bot!

fk0815:
There would be a 3D model here: https://grabcad.com/library/rfm95-lora-radio-module-1#! , but I'm not sure if that is allowed to be used.

carrotIndustries:
> There would be a 3D model here: https://grabcad.com/library/rfm95-lora-radio-module-1#! , but I'm not sure if that is allowed to be used.

Doesn't seem like it.

One last thing caught my eye: The Package is missing pin 1 indicators in silkscreen and assembly layer. Extend the line on the top left a bit and add a chamfer to the rectangle on the assembly layer.

fk0815:
Bot!

fk0815:
Bot!

PR #215, opened 2021-04-06
Add RT9193 LDO regulator

fk0815:


fk0815:
Bot!

fk0815:
Bot!

RX14:
The entity here might be worth renaming to be generic like the other LDO entity/unit/symbols. Also needs regulator tag and I'd remove the orientation-dependent texts.

fk0815:
Bot!

fk0815:
> The entity here might be worth renaming to be generic like the other LDO entity/unit/symbols.

The BP pin is somehow special for that regulator. It allows to bypass the internal voltage reference by a capacitor, which allows to operate the regulator as a very low noise device. Not sure if that should be kept as a generic entity/unit/symbol.

RX14:
There should probably be a `RT9193-GB`, which is inherited to `RT9193-33GB` as a 3.3V specific part. That means the other voltages can be added in the future easily.

fk0815:
Bot!

carrotIndustries:
For consistency reasons, it'd be good if the symbol had the same size as the "Buck converter with enable" one.

fk0815:
Bot!

PR #216, opened 2021-04-17
Fix a few small things in existing QFPs

fruchti:
Looking at #159 I noticed that the existing QFP packages had some minor issues. This PR is a quick fix for:

- Silkscreen refdes position near pin 1
- Silkscreen outline clearance
- Assembly layer refdes position and size
- Assembly layer pin 1 decoration size
- File and package names

Bot!


fruchti:
Bot!

PR #217, opened 2021-04-18
Add small version of the 2CH Feedthrough ESD Protection symbol

jue89:
I just removed the pin names. The symbol itself is verbose enough.


jue89:
Bot!

RX14:
Bot didn't do too well here oops :P 

I'd prefer not to overdo symbol variants, and decide on one vs another. In my opinion, this PR should modify the original symbol.

RX14:
Note that my 4ch ESD protection symbol doesn't have pin names, so this would be consistent with that. (it should probably have a box around it though) Perhaps redoing the whole symbol to be more consistent with both the discrete diode/schottky diode symbols and the 4ch ESD protection would be a good idea?

carrotIndustries:
Let's try this again:

Bot!

carrotIndustries:
Bot! again

carrotIndustries:
>  (it should probably have a box around it though) 

Good idea.

From my perspective, there's no reason for keeping the old symbol for this unit.

jue89:
Bot!

jue89:
I replaced the symbol as requested.

RX14 (approved):

fruchti:
Do we have a quasi-convention for in-symbol net junctions yet? I usually use the same size as schematic junction dots, but I using smaller ones (like you did here) inside symbols by default could make sense, too.

jue89:
Bot!

jue89:
> Do we have a quasi-convention for in-symbol net junctions yet? I usually use the same size as schematic junction dots, but I using smaller ones (like you did here) inside symbols by default could make sense, too.

I prefer the small dots since the diode symbols also has been reduced in size.

> Also, I’d prefer the diodes to be unfilled, like the other diodes in the pool.

Good point. Done.

RX14 (approved):

fruchti:
> > Do we have a quasi-convention for in-symbol net junctions yet? I usually use the same size as schematic junction dots, but I using smaller ones (like you did here) inside symbols by default could make sense, too.
> 
> I prefer the small dots since the diode symbols also has been reduced in size.

A smaller size also makes them easier to distinguish from regular junctions, so this is fine for me, too. I’d like to add a specific size to the convention, because there are different sizes in the pool right now:

- 0.4 mm (4ch ESD diode array)
- 0.35 mm (this PR)
- 0.25 mm (ZCD opto-TRIAC)

For reference: regular schematics junctions are 0.75 mm.

Personally, I find 0.25 mm a bit too small and 0.4 mm have the disadvantage that you have to change the grid to draw them. For the convention, I’d propose a size of either 0.35 mm like yours or 0.5 mm, which would amount to half the area of a regular junction dot. I’m fine with either, though. What do you think?

RX14:
0.35 sounds good to me

jue89:
0.35mm sounds good to me, too. It would be nice to have this information in the pool-convention.

carrotIndustries:
> 0.35mm sounds good to me, too. It would be nice to have this information in the pool-convention.
https://github.com/horizon-eda/horizon/commit/a1c776c0a627fae0c17ec71f9a0edfc771191aeb adds a tool for placing dots of the correct size in symbols

fruchti:
> [horizon-eda/horizon@a1c776c](https://github.com/horizon-eda/horizon/commit/a1c776c0a627fae0c17ec71f9a0edfc771191aeb) adds a tool for placing dots of the correct size in symbols

This is even better, thanks!

I added a rule to the convention in  horizon-eda/horizon-pool-convention@8b18e09.

PR #218, opened 2021-04-18
Add TPS56339 buck converter

RX14:
A bunch of SOT-23 buck converters use this pinout and layout, I've added only the TPS56339 for now


RX14:
Bot!

PR #219, opened 2021-04-18
Add molex picoblade 4-pin connector

RX14:
Manufactured and test soldered (by hand) on a board of mine

### DIFF
--- a/entities/connector/generic/004-mount.json
+++ b/entities/connector/generic/004-mount.json
@@ -1,0 +1,19 @@
+{
+    "gates": {
+        "a96a662e-2077-467f-9853-a01865f76ad6": {
+            "name": "Main",
+            "suffix": "",
+            "swap_group": 0,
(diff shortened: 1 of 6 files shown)


RX14:
Bot!

carrotIndustries:
The mounting pads should be called M1/M2 and should have electrical connection in case someone wants to connect them to a ground plane for increased rigidity.

RX14:
That's quite annoying since now it needs a different schematic symbol... But that seems sensible.

RX14:
Bot!

carrotIndustries:
Hm, I'd have used the generic 5 pin connector

RX14:
well its not really a pin, and it's super confusing to select a 4 pin connector and get a 5 pin schematic symbol with no explanation.

Connecting these pads to a plane makes a lot of sense mechanically, but it's hella awkward in the schematic whatever you do.

PR #221, opened 2021-04-21
Polishing Opto-Triac with Zero-Cross-Detection

jue89:
I've put some work into the Opto-Triac with Zero-Cross-Detection. I personally think this is the more beautiful version. Furthermore, no filled diodes anymore, as pointed out in #217.


jue89:
Bot!

RX14:
pretty!

PR #222, opened 2021-04-24
Add FTSH-107-xx-x-DV parts

RX14:
This header series is used by ST for their STDC14 header on ST-Link programmers. It's basically the standard arm 10 pin + a UART. The -A version has some alignment pins, so the package has NPTHs to accomodate them.


RX14:
Bot!

PR #224, opened 2021-04-28
Add Darlington array family ULN200xD

jue89:
* symbol, unit, entity: 7ch Darlington array with free wheeling diodes
* part: ULN200[1234]D1013TR by ST


jue89:
Bot!

RX14:
I think the pads in the package should all be 1.27mm part, not some 1.25 and some 1.3.

jue89:
Bot!

jue89:
Well I redid the package using the Footprint Generator ... and it happend again :D
Maybe @carrotIndustries has an idea, what's going wrong? It looks like small rounding errors.

jue89:
I corrected the pitch by hand. Looking at the diff clearly shows that the generator quantizes in 0.05mm steps.

jue89:
Bot!

carrotIndustries:
> Looking at the diff clearly shows that the generator quantizes in 0.05mm steps.

If you're using the IPC generator, that's a setting:

image

jue89:
Oh thank you for the info! I always thought, this "Round-off" setting is related to the "Rounded rectangular". What is the purpose of this setting? Why is its default set to `0.05 mm`?

RX14:
looks good to me now

jue89:
Bot!

jue89:
Enlarged junction diameter to 0.35mm to be aligned with #217.

carrotIndustries:
You could set the symbol pin driver to "open collector" so that the open-collectorness is visible from the symbol.

jue89:
Bot!

jue89:
I created the OC symbol by my-self. Otherwise it would have interrupted to net line.
Or do you prefer the original sign with the not-so-pretty interrupted net line?

PR #228, opened 2021-05-08
Add refdes and value text to single power symbol

jue89:
Without these, it's hard to say to which part the symbol belongs.

### DIFF
--- a/symbols/common/power-single.json
+++ b/symbols/common/power-single.json
@@ -1,11 +1,19 @@
 {
     "arcs": {},
+    "can_expand": false,
     "junctions": {},
     "lines": {},
     "name": "Single power",
     "pins": {
         "a8ba387d-1c68-4215-843f-9d669e48646d": {
+            "decoration": {
+                "clock": false,
+                "dot": false,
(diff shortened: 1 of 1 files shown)


jue89:
Bot!

carrotIndustries:
This symbol is intended to to directly on opamps and logic gates like so:

2021-05-08-125334_438x492_scrot

jue89:
Okay, I see. I haven't noticed this before.

Do you have any suggestions how to make this working with #227? Create an new symbol and unit for power? Create an alternative symbol? Let the user smash the symbol and remove refdes and value by hand?

RX14:
There's a few problems with the current solution, in addition to what @jue89 mentioned.

1. It's not discoverable. I had no idea that this power symbol was supposed to be used like that. If this remains as-is how do we teach users how this is supposed to be used?
2. Some parts have multiple opamps and a single power unit, it's weird that only one of the parts has power info.

carrotIndustries:
I thought that this concept was universally understood since it's done the same way in KiCad, EAGLE and probably other tools.

PR #230, opened 2021-05-09
Add KX023-1025

not7cd:
The KX023 is a robust, low-power, I2C/SPI, 3-axis accelerometer with integrated FIFO/FILO buffer that features a wide range of embedded functionality, including tap detection, orientation, activity, and wake-up algorithms.

Although not recommended for new design, there is quite big supply of them in well known pcb prototype supplier.

This could introduce first accelerometer to the pool, and LGA-16 package.

### DIFF
--- a/entities/ic/imu/KX023-1025.json
+++ b/entities/ic/imu/KX023-1025.json
@@ -1,0 +1,19 @@
+{
+    "gates": {
+        "0a106ba1-7f9a-4932-b729-5dcce1d09d73": {
+            "name": "Main",
+            "suffix": "",
+            "swap_group": 0,
+            "unit": "f0cb8d59-5e6a-4b57-af4a-896fc2a5a9f0"
+        }
+    },
+    "manufacturer": "Kionix",
+    "name": "KX023-1025",
+    "prefix": "U",
+    "tags": [
+        "ic",
+        "imu"
+    ],
+    "type": "entity",
+    "uuid": "ab6ec1b7-82c2-4d24-9397-027f52a61549"
+}
--- a/packages/ic/smd/lga/lga-16/package.json
+++ b/packages/ic/smd/lga/lga-16/package.json
@@ -1,0 +1,386 @@
+{
+    "arcs": {},
+    "default_model": "00000000-0000-0000-0000-000000000000",
+    "dimensions": {},
+    "junctions": {
+        "178ffe48-7ff4-4348-80eb-7a236a56b1de": {
+            "position": [
+                1000000,
+                -1550000
+            ]
+        },
+        "323d8097-4949-4c89-9f7f-96368c83a2f3": {
+            "position": [
+                1650000,
+                -1550000
+            ]
+        },
+        "3a441355-c511-46d0-8c14-869edb41205f": {
+            "position": [
+                -1650000,
+                -1550000
+            ]
+        },
+        "3ba4454b-677c-4b13-8b8a-16497dd6afca": {
+            "position": [
+                1650000,
+                1550000
+            ]
+        },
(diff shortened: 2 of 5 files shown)


carrotIndustries:
Bot!

RX14:
Thanks for the PR! A few things to improve:

- Fix the package checks (see [here](https://docs.horizon-eda.org/en/latest/create-package.html))
- There's no orientation marker for the silkscreen
- The description should be shorter, check a few other parts descriptions for ideas. Probably keeping just the first sentence will do here 
- I would keep `VDD` uppercase.
- `SCLK` and `SCL` and the other I2C/SPI dual purpose pins should be represented with alternate pin names. The alternate pin names as used in this PR are not used in the datasheet so shouldn't be used.
- REFDES and VALUE should be placed to the right of the power pins on the top edge so that a long REFDES doesn't interfere
- `nCS` should be changed to `~CS` so that it shows with an overbar.

RX14:
Bot!

RX14:
Looks like the package checks are still failing :(

The best way to create a valid courtyard is with the "generate courtyard" tool which does it all for you. You'll need to delete the existing courtyard to get the courtyard generator to run.

For the silkscreen I'd suggest starting from what the silkskreen generator gives you and editing it to add the pin1 marker (the style where you delete the line by pin1 is correct here)

RX14:
Simple box-shaped parts like this shouldn't use orientation-dependent text placements either, that's designed for stuff like resistor symbols where the text placement changes significantly depending on rotation.

carrotIndustries:
> Simple box-shaped parts like this shouldn't use orientation-dependent text placements either

I've thought about adding a check for this that generates a warning if orientation-dependent texts are set and the symbol has more than 4 pins. Can you think of any cases where this check would be a false positive?

RX14:
> I've thought about adding a check for this that generates a warning if orientation-dependent texts are set and the symbol has more than 4 pins. Can you think of any cases where this check would be a false positive?

Would it be possible to check for simple box-shaped parts instead? i.e. 1 poly with 4 vertices?

carrotIndustries:
> Would it be possible to check for simple box-shaped parts instead

We already have that, it's what generates the "is box symbol" result.

RX14:
Right. I think your suggestion is fine then.

carrotIndustries:
https://github.com/horizon-eda/horizon/commit/f421c9b19223bd140156eac8f8f46bf5ae192277 adds this warning if there are more than 4 pins:

2021-05-15-170835_969x723_scrot

RX14:
Might need to be a bit more obvious, for a new user that message's wording isn't clear that this message applies to them

carrotIndustries:
> Might need to be a bit more obvious, for a new user that message's wording isn't clear that this message applies to them

Hm, how about: "Don't apply orientation-specific texts to this symbol as it has more than 4 pins. Only use orientation-specific placements for small symbols that are expected to be used in multiple orientations."?

not7cd:
Anyway, I will address those issues later, maybe add LGA-8 and other 
IMU's from Kionix lineup.

Also there are guidelines for designing footprints. Mine are much bigger 
when it comes to copper layer. I'm unsure how it will behave in 
production. I will know in 2 weeks from now.

https://kionixfs.azureedge.net/en/document/TN009-PCB-Design-Guidelines-for-3x3-LGA-Sensors.pdf

Another thing is 3D model. I'm using QFN 3x3mm as it fits the package 
size, but pins are mismatched. Could I create a generic 3x3mm model, 
without pins modelled in?

RX14:
> Hm, how about: "Don't apply orientation-specific texts to this symbol as it has more than 4 pins. Only use orientation-specific placements for small symbols that are expected to be used in multiple orientations."?

perfect!

PR #231, opened 2021-05-23
Molex Minifit-Jr 5566

moritz-meier:
Molex Minifit Power Connector (ATX Connector) 5566 Series Vertical THT with PCB Pegs

12 Packages
12 Base Parts + 170 Parts
All Variants (Plating, Flammability, Drain Hole) as separate Parts

Any suggestions on the package or the part structure?


moritz-meier:
!Bot

moritz-meier:
Bot!

moritz-meier:
Added other Variants, but only those which are known to Octopart (some part numbers from the datasheet are unknown, so i think these parts are unnecessary)

Regarding the Warning on the Package: The package outline is a closed polygon, but I used a single line to close the rectangle from the retention mechanism at the top. Should I remove this line?

Regrading the Error on the Package: I am not sure whats the problem here? Of cause the Package is placed on the Silkscreen?

moritz-meier:
Bot!

PR #233, opened 2021-06-05
Update QFN-32 package and add hand-solderable version

RX14:
This is basically the same as the old package, but with roundrects instead of half-obround, silkscreen to the new guidelines, and a name change to distinguish this version of qfn-32 from others with different pitches and exposed pads size. Also adds a hand-solderable version with more pad length exposed.

This rework was precipitated by creating another version for an MSP430 package which had a smaller exposed pad (2.1mm)


RX14:
Bot!

RX14:
Bump

PR #234, opened 2021-07-13
Adding NPN-BJT Nexperia MMBT3904

atoav:
This adds the Nexperia MMBT3904 BJT. It uses the same SOT-23 footprint as #45


atoav:
Bot!

atoav:
Bot!

RX14:
We have SOT-23 already, no?

atoav:
Yeah, this is one of those that would be really cool to have for the package alone : )

RX14:
- Since NPN/PNP transistors in sot-23 package have standardized pin layout (i think?), it might be a good idea to make a base part
- SOT-23 comments from #45 applies.

atoav:
> Since NPN/PNP transistors in sot-23 package have standardized pin layout (i think?), it might be a good idea to make a base part

I wasn't too sure about this, which is why I left it out

carrotIndustries:
Apparently other pinouts do exist: https://www.eevblog.com/forum/beginners/do-they-exist-sot-23-pnp-transistors-with-non-standard-pinout/

Anyhow, I think adding generic base parts for the happy path is a good idea as it makes swapping parts less error prone.

PR #235, opened 2021-07-13
Adding Barrel Connector Unit/Symbol/Entity, Würth 6941xx301002 package and part

atoav:
This adds a BOSS guitar pedal-style power connector. There are many no-name parts that use the same footprint, but I found a propper one at Würth: https://www.we-online.de/katalog/datasheet/6941xx301002.pdf

The 3D model comes from Sparkfun, see commit message ee76858 for more details on license

Note: This is the cleaned up version of #148


atoav:
Bot!

atoav:
Not sure what the warning on the symbol means, but in my eyes this one is ready (also used in production already so I am quite sure that it works)

RX14:
Oh, I was about to PR this myself but without the 3D model, thanks! This is basically identical to what I made, so thumbs up from me!

PR #236, opened 2021-08-01
Adding BC85x SOT-23 PNP transistors

atoav:
This PR adds Nexperia BC85x PNP transistors as seen here: https://assets.nexperia.com/documents/data-sheet/BC856_BC857_BC858.pdf

The part uses the SOT-23 package from #45


atoav:
Bot!

carrotIndustries:
Bot! (again, sot-23 package should be gone now)

carrotIndustries:
The A/B/C variants shouldn't be listed as orderable MPNs as they differ in current gain. Orderable MPNs should only be used if the parts only differ in packaging (e.g. tape&reel vs. tube), but are actually the same actual part.

PR #237, opened 2021-08-04
Fix SOT-363 assembly outline layer

RX14:
The assembly outline for SOT-363 was on the package layer, not the assembly layer.

I also converted the pads to roundrect while I was fixing this package.


carrotIndustries:
Bot!

PR #241, opened 2021-11-14
Add ADR44x Voltage
References

fk0815:
New Part with symbol, unit and entity
Ultralow Noise, LDO XFET Voltage References with Current Sink and Source


fk0815:
Bot!

carrotIndustries:
- GND pin text should be perpendicular
- The parts listed in the orderable MPNs are different parts, these should be derived parts. Only use oderable MPNs for different packaging variants (e.g. tube vs tray).

fk0815:
Bot!

fk0815:
Bot!

PR #242, opened 2021-11-14
Add ADuM1300 Digital Isolator

fk0815:
Add part with symbol, entity and unit, icluding SOIC-16W package with 3D model
Triple-Channel Digital Isolators


fk0815:
Bot!

carrotIndustries:
These should be different parts, only ADuM1300ARW and ADuM1300ARW-RL should be listed as orderable MPNs as these are the same part, just different packaging options,

fk0815:
Bot!

fk0815:
Bot!

carrotIndustries:
I'd prefer there being a specific base part rather than using a random variant.

fk0815:
Not sure if I understand correct, but do you mean to change this:

image

in a way so that there is a ADuM1300 part and the ADuM1300[ABC]RWZ should be derived parts from that?
In such a case the ADuM1300 part would be not orderable, can you name an example for that?

fk0815:
Bot!

carrotIndustries:
>  In such a case the ADuM1300 part would be not orderable, can you name an example for that?

Yes, that's how it's supposed to be, see the STM32F103RxTx for an example. It's recommended to set the 'base part' flag for parts like these.

PR #243, opened 2021-11-16
Add STM32L412CxTx device

fk0815:
STM32L412 with QFP case


fk0815:
Bot!

carrotIndustries:
Unit could use some alternate names. See https://github.com/horizon-eda/horizon/blob/master/scripts/stm32_to_json.py

fk0815:
I do not understand what is missing. For the new part I did not add a new unit, only a package and a part that it using it.

E.g. compare pin PA13 between the current alternative pin functions:
image
compared with the import of the json data generated with stm32_to_json.py with the latest db/mcu/STM32L412CBTxP.xml:
image
the pin name is different. Should I change that for the existing unit?

carrotIndustries:
> I do not understand what is missing. For the new part I did not add a new unit, only a package and a part that it using it.

Sorry, my bad.

PR #250, opened 2021-12-07
Make SMD-3225 (3.2x2.5mm) package generic, add YSX321SL

unrelentingtech:
This package is not proprietary to a manufacturer so move it to generic crystal packages like SMD-2520 already is. Add one more part using it.


unrelentingtech:
argh, had to add the deletion manually, the GitHub integration doesn't seem to work well with "Move/rename". (Also after "Move/rename" that `"_imp"` subobject in the json got deleted…??)

carrotIndustries:
Bot!

RX14:
Orderable MPNs are designed for use where the only difference between parts are packaging. There should be a base part for YSX321SL and then inherited parts for each frequency and grade. You don't have to list every part in the range in this PR either.

RX14:
Bot!

RX14 (approved):

carrotIndustries:
@RX14 same, as https://github.com/horizon-eda/horizon-pool/pull/252#issuecomment-994147902, what about repeating the MPN in the orderable MPNs?

RX14:
Bot!

RX14:
Somehow the old part file got re-added, and needs to be removed to avoid a duplicate in the pool.

unrelentingtech:
Bot!

RX14:
I think the 3d model should be renamed to be generic for the package now, other than that this is perfect from my side.

unrelentingtech:
There actually was no 3D model, whoever worked on the footprint before committed a path to a local model that's not in the repo… it's even in the path: `3d_models/local/crystal/NX3225SA v2.step`.

Added the KiCad one.

RX14:
Bot!

PR #251, opened 2021-12-08
Add STM32L15xCxTx

unrelentingtech:
I'm not sure if putting an 'x' in the number is a good idea but the [datasheet](https://www.st.com/resource/en/datasheet/stm32l151c8.pdf) is shared between STM32L151x6/8/B and STM32L152x6/8/B, the pinout figure is titled "STM32L15xCx LQFP48 pinout".

Also I guess other STM32s all have short pin names but it does seem like e.g. "WKUP2" is part of the whole "PC13-WKUP2" name, it was imported like that from the CubeMX xml file and this is how it is in the pinout in the datasheet as well… hmm.

### DIFF
--- a/entities/ic/mcu/stm/STM32L15xCxTx.json
+++ b/entities/ic/mcu/stm/STM32L15xCxTx.json
@@ -1,0 +1,21 @@
+{
+    "gates": {
+        "ac6d19cf-b1bd-4c94-b526-095cbac224f4": {
+            "name": "Main",
+            "suffix": "",
+            "swap_group": 0,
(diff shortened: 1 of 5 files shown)


RX14:
bot!

carrotIndustries:
- Symbol should be expandable
 - Don't use orderable MPNs for parts that differ in memory size, these should be distinct parts. Only use orderable MPNs for variation in packaging such as tape vs tray.

RX14:
Bot!

RX14:
Should mark the part as a base part and add the specific part number you were using.

RX14 (approved):

PR #252, opened 2021-12-08
Add USB connector TYPE-C-31-M-12 from Korean Hroparts

unrelentingtech:
This is a USB 2.0 connector (no superspeed pins) from the same manufacturer as TYPE-C-31-M-30.
There are two versions of the datasheet: [the one on LCSC](https://datasheet.lcsc.com/lcsc/1811131825_Korean-Hroparts-Elec-TYPE-C-31-M-12_C165948.pdf) and [the one on the vendor's website](http://www.krhro.com/uploads/soft/180320/1-1P320120243.pdf) — the latter has a more detailed "recommended PCB layout" but the PDF is rotated (lol).

The footprint is kinda similar to DX07S016JA1R1500 but not really, so I've added one based on a KiCad import. The big pads are for two pins each (at least on the actual USB-C side) so I've called them "A1B12"/etc. in the footprint, I'm not sure what the best name would be (picking one of two feels ehhhhh not great, just inventing something like "VBUS1" "VBUS2" also I'm not sure about).

There is a 3D model [here](https://grabcad.com/library/type-c-31-m-12-1) but it's not explicitly licensed under compatible terms :/ (GrabCAD by default applies [a non-commercial-only license](https://grabcad.com/questions/what-type-of-license-governs-use-reuse-modifications-and-distribution-of-the-cad-files-on-grabcad))

### DIFF
--- a/packages/connector/usb/krhro/type-c-31-m-12/package.json
+++ b/packages/connector/usb/krhro/type-c-31-m-12/package.json
@@ -1,0 +1,626 @@
+{
+    "arcs": {},
+    "default_model": "00000000-0000-0000-0000-000000000000",
+    "dimensions": {},
+    "grid_settings": {
+        "current": {
+            "mode": "square",
+            "name": "",
+            "origin": [
+                0,
+                0
+            ],
+            "spacing_rect": [
+                1000000,
+                1000000
+            ],
+            "spacing_square": 100000
+        },
+        "grids": {}
+    },
+    "junctions": {
+        "0d37d21e-5f8c-45a8-b8ed-991500fe086c": {
+            "position": [
+                -3900000,
+                4800000
+            ]
+        },
+        "19252aca-d355-4d17-8dc8-9b7f9d734b4c": {
+            "position": [
+                -4900000,
+                4300000
+            ]
+        },
+        "22ecc6df-fbda-42a3-b5d2-3b5fc0ab87b9": {
+            "position": [
+                -4900000,
+                4800000
+            ]
+        },
+        "303c2741-f0c9-483b-9cc6-1334423ad1f5": {
+            "position": [
+                -4800000,
+                0
+            ]
+        },
+        "35788b56-9a6a-4d11-882e-aab5d992df1b": {
+            "position": [
+                3900000,
+                4800000
+            ]
+        },
+        "3a62caa6-f2ed-4923-9fcf-d1efab9f5d24": {
+            "position": [
+                4800000,
+                -3600000
+            ]
+        },
+        "3f653e1c-12a6-4373-9450-b1357b380959": {
+            "position": [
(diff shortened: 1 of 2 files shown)


RX14:
Bot!

RX14:
Interestingly there seem to be a family of parts with similar footprints, I highly suspect them to be close enough to be footprint-compatible (despite the manufacturers recommending slightly different footprints). One that I have a reel of is the Shenzhen Kinghelm KH-TYPE-C-16P-T.

Given these similarities, it's possible that these parts might be designed to be compatible with another part by a more well-known manufacturer. Either way, it might make good sense to use one package in horizon to cover all these parts. Thoughts?

unrelentingtech:
Yeah, that one is very likely to be compatible. We don't know which is the "more well known" one (well, kicad-footprints has this one) so I guess one of them would just use the other's footprint?

jue89:
#196 looks quite similar ... but I haven't checked the dimensions, yet.

unrelentingtech:
Yeah, I mentioned this in the original post:

> The footprint is kinda similar to DX07S016JA1R1500 but not really

And by "not really" I mean that e.g. the through-hole pins are significantly lower on this one.

jue89:
> Yeah, I mentioned this in the original post:
> 
> > The footprint is kinda similar to DX07S016JA1R1500 but not really
> 
> And by "not really" I mean that e.g. the through-hole pins are significantly lower on this one.

Ah I see. Sry, haven't realized that!

jue89:
First review:
- [x] Part: Pin and Pads are correctly assigned
- [x] Package: Dimensions are correct
- [ ] Part: Please don't repeat the MPN in the orderable MPNs - just leave it empty

RX14:
> And by "not really" I mean that e.g. the through-hole pins are significantly lower on this one.

There's a minefield of parts which look compatible but some are and some are not :(

carrotIndustries:
I remember having seen that one somewhere... https://twitter.com/cursedconnector/status/1460698610690105345

RX14:
> @RX14 do you recall anything related to that?

nope :(

Though orderable MPN can repeat MPN if there's more than one orderable MPN

RX14:
> I remember having seen that one somewhere... https://twitter.com/cursedconnector/status/1460698610690105345

Oof, looks like we need a varification on the pinout before this is added to the library?

unrelentingtech:
Hopefully that's already been done with the KiCad one that's imported here, but let's wait for a board I've used this in to arrive…

unrelentingtech:
TYPE-C-31-M-12

[Verified](https://github.com/unrelentingtech/ferrispark) :)

RX14 (approved):

RX14:
lets merge this!

PR #255, opened 2021-12-12
Add Single Buffer SN74AHCT1G125 for Level Shifting

jue89:
This PR adds a generic unit, symbol, entity and the vendor specific part.

### DIFF
--- a/entities/ic/buffer/1ch-buffer-with-en.json
+++ b/entities/ic/buffer/1ch-buffer-with-en.json
@@ -1,0 +1,20 @@
+{
+    "gates": {
+        "da749b97-3194-4a93-9db8-aa9b019b4443": {
+            "name": "Main",
+            "suffix": "",
+            "swap_group": 0,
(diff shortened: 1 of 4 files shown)


jue89:
Bot!

jue89:
Bot!

RX14:
Back [a while ago](https://github.com/horizon-eda/horizon-pool/pull/214#issuecomment-818284617) we decided that inverted pin decorations should only be used on logic gates, not ICs. Since the inverted decoration is on the box, not the gate here, I think it should be `~OE` instead.

jue89:
> Back [a while ago](https://github.com/horizon-eda/horizon-pool/pull/214#issuecomment-818284617) we decided that inverted pin decorations should only be used on logic gates, not ICs. Since the inverted decoration is on the box, not the gate here, I think it should be `~OE` instead.

Oh I haven't noticed that. But I recall a discussion about that in the past :D Sorry, I didn't remember it.

I'll fix it tonight. Thank you for looking into this PR :)

jue89:
Bot!

PR #258, opened 2021-12-21
Add rotary encoder / Bourns PEC12R-4xxF-Sxxxx

unrelentingtech:
I have ordered an "EC12SXB2" which as it [turns out](https://community.volumio.org/t/plugin-rotaryencoder-ii/48538/14) is an exact clone of the [Bourns PEC12R](https://www.bourns.com/docs/product-datasheets/pec12r.pdf)-4217F-S0024. (Should clones like this be added inheriting the original? This one doesn't even have a vendor name lol)

Footprint imported from [this](https://github.com/windsorschmidt/kicad-library-win/blob/master/win.pretty/PEC12R-4xxF-Sxxxx.kicad_mod), symbols from the default KiCad library.

(The importer does not support `np_thru_hole roundrect` holes! I have recreated the mounting holes manually.)

BTW the 3xxF models have the same footprint, but 2xxF do not. So unfortunately I couldn't invent good common name, it would have to be something ugly like `(3|4)xxF` :D


frmdstryr:
Bot!

RX14:
I'd like to add the `PEC12R-4xxF-Nxxxx` part which uses the no switch symbol/entity. The parts should also be marked as base parts, since there are multiple part numbers which descend from it. You should probably ass the specific part you used as a child of that part too.

unrelentingtech:
done

RX14:
Bot!

RX14 (approved):

carrotIndustries:
Is there a rendering bug with the bot or what is this symbol supposed to represent?

Symbol

unrelentingtech:
Well that's a question for the original kicad author…

I think the three lines and the angle things represent the rotary knob. Kicad users love fancy symbols, you've seen the standard library's SD card right?

unrelentingtech:
Bot!

- updated symbols to look like [this](https://electronics.stackexchange.com/questions/493221/number-of-pins-of-elements) (I guess something like this actually was in kicad, the arcs just didn't import)
- added 3D models [from the vendor](https://www.bourns.com/resources/design-tools/design-files) (!!!)
  - haven't found explicit copyright terms but this is intended as a resource for board design, 
- adjusted the side holes based on 3D models

RX14:
> * adjusted the side holes based on 3D models

I'd always trust a schematic footprint over a 3d model.

I'm also extremely wary of distributing any 3d models with unkown licence terms.

frmdstryr:
I created a 3d model from the datasheet and physical "EC11" clone/variant, it looks like the slots will not line up. I think the slots should stay at the original 5.6mm spacing of the datasheet.  

Using a caliper the inside measurement on those mount pins is ~9.8mm and the outside is around 11.4mm. Idk what the actual Bourn's part measures.

image

image

unrelentingtech:
@frmdstryr very nice! However EC11 is different from EC12. My EC12 clones arrived recently too, the inner measurement is larger than your outer one:



On EC12 these pins extend from the *top* part of the metal thingy (so there are like three strips of metal on the side where your model has two).

Either way these are very bendable :)

frmdstryr:
I see, sounds like they need to be different packages then, what a waste....

unrelentingtech:
or we could compromise on a middle position such that both would require a tiny bit of bending :D

BTW, what software did you use for 3D modeling? Would be nice to have the models for these variations derive from a common source.

frmdstryr:
> what software did you use for 3D modeling?

I use my own... "DeclaraCAD"

The EC11 models are at codelv.com/assets/models/ + ec11-rotary-encoder-w-switch.step , another with 'no-switch' with the pins removed. Same license as KiCad (CC BY-SA 4.0) so they can be added to the pool.

unrelentingtech:
Bot!

rebased, removed 3d model

carrotIndustries:
The symbol looks a bit odd, might be due to the kicad symbol importer not supporting arcs.

PR #260, opened 2021-12-29
Add Horizon logo silkscreen decal

unrelentingtech:
Converted from SVG via Inkscape's DXF export. Probably should've downscaled in Inkscape first, the circles ended up *quite* detailed… although the renderer can handle it just fine :)

### DIFF
--- a/decals/horizon-silk.json
+++ b/decals/horizon-silk.json
@@ -1,0 +1,431 @@
+{
+    "arcs": {
+        "31331b42-c31d-49fb-a8c0-d087060cdeaf": {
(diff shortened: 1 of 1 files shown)


unrelentingtech:
Bot!

carrotIndustries:
Bot! again

carrotIndustries:
You could significantly reduce the vertex count by using arcs, the outside circle could be done by a line circle.

unrelentingtech:
Oh right, that works nicely. Did both using line circles actually.

Bot!

unrelentingtech:
Bot!

// Oh right, that works nicely. Did both using line circles actually.

PR #264, opened 2022-01-23
Add TI LDC1x1x inductance to digital converter ICs

unrelentingtech:
https://www.ti.com/product/LDC1612 this kind of thing. (The first 'x' in the model number only changes bit depth, the second one is the number of channels.)

Footprints are imported from kicad, WSON-12 (LDC1612) I have actually tested on a real board.


unrelentingtech:
Bot!

carrotIndustries:
- GND/VCC pin names should be perpendicular
 - assy refdes should fit within assy outline
 - silkscreen refdes should be roughly above package

PR #265, opened 2022-01-27
Created ANSI A & B frames.

nellump:
Based on existing ISO A4 and A3 frames, respectively.


carrotIndustries:
Bot! now does frames as well

PR #272, opened 2022-02-06
Add XC6802 charget IC

fk0815:
SOT-25 is an alias for SOT-23-5


fk0815:
Bot!

fk0815:
Bot!

carrotIndustries:
- Symbol can be made a bit more compact
 - I don't think a generic entity/unit/symbol makes sense here. If we find out that there are other parts that have the same pins, we can always make them generic.

fk0815:
Bot!

carrotIndustries:
- entity should not have generic tag

fk0815:
Bot!

PR #282, opened 2022-02-17
Adding IDC Box headers

atoav:
This commit adds generic shrouded IDC box headers with 2.54mm pin spacing.

More specifically:
- all variants from 2×3 pins to 2×50 pins. I am not sure if all of those are actually available.
- the existing generic connector entities/units have been reused
- generated via script, so quick changes/additions/different pitches are possible

image

The 3D models are generated via cadquery because I didn't found models at kicad.


atoav:
Bot!

atoav:
@carrotIndustries It seems like the bot failed without actually delivering a report if I judge the situation rightly?

carrotIndustries:
From the actions:

>      "message": "Body is too long (maximum is 65536 characters)"

There are already some similar parts in the pool, such as "Generic IDC header 2×3, 2.54mm pitch, vertical" These were generated by https://github.com/horizon-eda/horizon/blob/master/src/gen-pkg/gen-pkg.cpp with the 3D models copied from KiCad "Connector_IDC.3dshapes/IDC-Header_2x03_P2.54mm_Vertical.step"

atoav:
> There are already some similar parts in the pool, such as "Generic IDC header 2×3, 2.54mm pitch, vertical"

Yes I am aware of those – I used them for many projects as of now. The shrouded/keyed box headers however have a significantly bigger footprint, and checking for potential collisions is certainly one of the added values of an EDA application, which was my main rationale for adding those parts. As far as I am aware these box headers are still very common in internal PCB-to-PCB connections using ribbon cables. The one advantage they have over "normal" unshrouded IDC headers is that you cannot connect them 180 degrees wrong.

It is 360 MB of step files tho, maybe we need to find a different way to deal with that 3D data in the long run?

RX14:
I'm ambivalent to having these in the main pool myself but there were separated pools for generated components like this before. I think it makes sense.

carrotIndustries:
>  The shrouded/keyed box headers however have a significantly bigger footprint

Perhaps I'm overlooking something but to me they look pretty much identical to the existing ones:

image

image

atoav:
Shame on me, I didn't find that one. I probably have been too distracted by the other Generic Pin header and searching for "box header" didn't bring anything up.

How shall we proceed? I would at least need to add 2×5 as well, because this is one of those I need. Should I create a new PR for that one where I try to match the format of the other ones?

Or would it be appropriate to change "IDC header" to "IDC Box Header" in the names so people who know it by Box header like me can find/differenciate them easier?

Any way, this PR can be closed for now : )

carrotIndustries:
> How shall we proceed? I would at least need to add 2×5 as well, 

We have that one, it's called "IDC header 2×5, 2.54mm pitch, vertical"

PR #283, opened 2022-02-17
Add MCP23017-E/SP I/O expander

unrelentingtech:
Fully tested in my split keyboard. Only the DIP part here, but the symbol can be reused for the other package variants.

### DIFF
--- a/entities/ic/gpio/microchip/MCP23017.json
+++ b/entities/ic/gpio/microchip/MCP23017.json
@@ -1,0 +1,19 @@
+{
+    "gates": {
+        "f2352505-89d1-407a-8566-a7afb0976c99": {
+            "name": "Main",
+            "suffix": "",
+            "swap_group": 0,
+            "unit": "5ca5c273-35c6-4119-b683-c83d1811862d"
+        }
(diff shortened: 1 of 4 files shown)


carrotIndustries:
- Pins are too long and power supply pins should use perpendicular text
- SDA should be bidirectional
- Symbol could be a bit more narrow
- Why is it expandable?

unrelentingtech:
Bot! // fixed. I guess it was expandable just because it's a big IC with a bunch of GPIO which looks microcontroller-like enough for me. Or something.

carrotIndustries:
- Pins are still too long

unrelentingtech:
Bot!

carrotIndustries:
- Refdes/Value should be on the right side of the VDD/VSS pins so that the text won't collide with them.

unrelentingtech:
Bot! // hmm right, that would make long text possible, but the alignment doesn't look as nice now…

RX14 (approved):

unrelentingtech:
squashed the commits, no actual changes since the bot pass

Issue #285, opened 2022-02-19
DIP packages have inconsistent pad hole sizes


nellump:
Pads' hole/ring sizes vary from 0.700/1.300mm to 0.850/1.350mm, with stops in between.  All of the packages I looked at feature the same pattern.  (Probably an error that propagated via duplication.)

nellump:
Fixed in PR #286.  (Assuming 0.750mm holes and 1.350mm pads are suitable.)

PR #286, opened 2022-02-20
Set all DIP packages to 0.750mm hole, 1.350mm pad.

nellump:
Reference issue #285.  DIP packages were edited to use holes of 0.750mm and pads of 1.350mm, as necessary.

### DIFF
--- a/packages/ic/th/dip/direct_solder/DIP10_7.62mm_lead_span_2.54mm_pitch/package.json
+++ b/packages/ic/th/dip/direct_solder/DIP10_7.62mm_lead_span_2.54mm_pitch/package.json
@@ -250,7 +250,7 @@
             "name": "9",
             "padstack": "296cf69b-9d53-45e4-aaab-4aedf4087d3a",
             "parameter_set": {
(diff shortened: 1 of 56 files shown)


RX14:
Bot!

RX14 (approved):

carrotIndustries:
So is that the size that other people assume for dip packages?

nellump:
The first five pads of all of the DIP parts in horizon-pool are set to
0.750/1.350mm hole/pad size, which seemed reasonable.  35 mil is a
typical hole diameter for larger PTH parts.  FWIW, KiCad uses
0.800/1.600mm for their DIP footprints.  Is there an IPC standard for
these with which we should aspire to comply?

nellump:
> Is there an IPC standard for these with which we should aspire to comply?

The Internet tells me the applicable standard is JEDEC MS-001, which specifies an 18 mil pin width.  A quick spot check with parts and boards on hand shows that a DIP-14 pin easily fits into a 30 mil hole.  An 0.800mm (~31 mil) hole sounds reasonable.

PR #294, opened 2022-07-05
Adding Instrumentation Amplifiers by Analog Devices

atoav:
This adds multiple instrumentation amplifiers by Analog Devices, as well as a unit, entity and symbol for similar instrumentation amplifiers with a gain resistor and a reference output.

All DIP/SOIC instrumentation amplifiers by AD have been added, parts where the package was missing in the official pool (e.g. MSOP) have been skipped.

### DIFF
--- a/entities/ic/manufacturer/ad/AD620.json
+++ b/entities/ic/manufacturer/ad/AD620.json
@@ -1,0 +1,26 @@
+{
+    "gates": {
+        "6a103702-c217-455e-99c6-3a782da60008": {
+            "name": "Main",
+            "suffix": "",
+            "swap_group": 0,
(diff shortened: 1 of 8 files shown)


atoav:
Bot!

atoav:
Two comments on this:

- Bot says "x Gate suffix "" doesn't match regex" – the help text in horizon for the gate field tells you that for single gate entities you may use an empty string for the gate. So either this help text is outdated, or the rule should not apply for parts with one gate and power pins. I know I have a power gate here, should I still use "A" as a Gate suffix?
- Pin Ref is not on bounding box, but this is intended.

atoav:
Bot!

PR #295, opened 2022-07-23
Add orientation specific text alignment for MOSFETs

oxplot:
Screenshot from 2022-07-24 00-53-58
Screenshot from 2022-07-24 00-54-14


carrotIndustries:
Bot!

PR #296, opened 2022-07-23
Fix the orientation of text in MOSFET symbols

oxplot:
After looking at the screenshots of the previous PR for a bit, I realized that I probably needed to fix the orientation of other text as well — so here it is.
Screenshot from 2022-07-24 01-37-50
Screenshot from 2022-07-24 01-38-01


oxplot:
@carrotIndustries is this good to go?

PR #299, opened 2022-08-31
Adding TO-220-3 packages and IRFZ 44N Mosfet

atoav:
This commit adds:
- TO-220-3 (vertical) package
- TO-220-3 (tab-down) package
- Infineon IRFZ44NPBF Mosfet

The 3D models are `TO-220-3_Vertical.step` and `TO-220-3_Horizontal_TabDown.step` from KiCAD: https://kicad.github.io/packages3d/Package_TO_SOT_THT


RX14:
This looks good! Only two things I would consider: I prefer kicad's idea of the silkscreens for this part. The silk should more closely hug the legs on the horizontal package, and there should be a fairly clear indication of which side the heatsink goes from the silkscreen instead of using an arrow towards the gate.

In addition, perhaps consider removing the "pbf" from the MPN and putting it in orderable MPNs. It's electrically irrelevant and just makes the MPN on the schematic longer for no gain.

carrotIndustries:
> In addition, perhaps consider removing the "pbf" from the MPN and putting it in orderable MPNs. It's electrically irrelevant and just makes the MPN on the schematic longer for no gain.

I get the point, but that MPN including PBF appears to be the one and only MPN for this part, so there are no alternate orderable MPNs.

atoav:
> Only two things I would consider: I prefer kicad's idea of the silkscreens for this part. The silk should more closely hug the legs on the horizontal package, and there should be a fairly clear indication of which side the heatsink goes from the silkscreen instead of using an arrow towards the gate.

You have a picture of that? I am not using KiCad

PR #300, opened 2022-08-31
Update 603 LED package to rounded pads and fix silkscreen clearance

frmdstryr:
Change to rounded pads and adjust the silkscreen spacing so it does not fail the default clearance checks.


frmdstryr:
Bot!

carrotIndustries:
I don't think we need to have 0603 in the tags.

PR #302, opened 2022-10-23
TL074 Opamps: Move parts to orderable MPNs

atoav:
This PR cleans up the existing TL074 parts:

- adds a datasheet URL to the former base part
- removes children parts and adds them as orderable MPNs
- renames the MPN on the base part
- renames the files for the base part


atoav:
Bot!

RX14:
Could you remove the query parameter from the datasheet URL?

PR #304, opened 2022-10-26
Adds packages, base parts 3D models for 6.3mm Resistors

atoav:
As proposed by @carrotIndustries [here](https://github.com/horizon-eda/horizon-pool/pull/48#issuecomment-1291209049) I added the packages, 3D-models and base parts for 6.3mm through hole resistors (both horizontal and vertical orientations).

The base parts are actually altered variants of the ones I am using in [my repo](https://github.com/atoav/horizon-pool-yageo-th-resistors), as they have been including some Yageo-specific details. I will add these new base parts as the bases of my resistors-bases as well. To make them compatible with the official pool.

The 3D-models (as mentioned already in the original PR #48) have been modelled by myself. I was toying with the idea of writing a script to add colored stripes, but reached the conclusion that the amount of data this would add would not be worth it. People who want that kind of detail in their 3D-models are better served with using textures and Blender anyways.


RX14:
Bot!

atoav:
Btw. the silkscreen error is expected here. The TH part goes on top of the silkscreen. This error will also happen with the TO-220-3 and similar packages.

RX14:
The same physical resistor is used for both horizontal and vertical parts, surely this should be a single base part with 2 package variations?

atoav:
Now that you mention it, it makes a lot of sense. These packages have been very old, so it might be that back then the alternate package mechanism was not implemented yet (or, equally likely: it *was* implemented, but I did not know about it back then). Anyways I fixed it here, and I will fix it in my repo as well.

RX14:
Bot!

RX14:
- Reference Designators should only be on the silkscreen and assembly layer
- Assembly layer refdes should ahve a small enough font size to fit `R1234` comfortably, the current text doesn't
- The vertical package assembly layer refdes is a bit small, should probably be vertical

PR #262, opened 2022-01-09
Add CC7V-T1A 32768Hz crystal

fk0815:
Add package and part

PR #305, opened 2022-11-11
Add NX3215SE Crytal

fk0815:
Same Package as CC7V only part is new


carrotIndustries:
Bot!

carrotIndustries:
> Same Package as CC7V only part is new

So why is there a new package in this PR?

fk0815:
> > Same Package as CC7V only part is new
> 
> So why is there a new package in this PR?

I think that is because of this PR is not merged yet: https://github.com/horizon-eda/horizon-pool/pull/262

PR #306, opened 2022-11-13
Adds E-Switch EG2201 vertical DPDT slide switch

atoav:
I needed a vertical DPDT slide switch and added this one by E-Switch: https://sten-eswitch-13110800-production.s3.amazonaws.com/system/asset/product_line/data_sheet/119/EG.pdf

### DIFF
--- a/entities/mechanical/switch/switch_dpdt_on-on.json
+++ b/entities/mechanical/switch/switch_dpdt_on-on.json
@@ -1,0 +1,20 @@
+{
+    "gates": {
+        "2e41400f-138b-4ab4-9ba7-c3255ff0adbd": {
+            "name": "Main",
+            "suffix": "",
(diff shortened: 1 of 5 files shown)


carrotIndustries:
Bot!

carrotIndustries:
- variants should not be orderable MPNs as they differ in actuator length
 - one symbol pin is shorter than the other
 - value should be blank
 - mechanical pads should just be M1 .. M4
 - we don't have a layer legend, but what's the blue line near MC1?

atoav:
Bot!

atoav:
@carrotIndustries I fixed all the things except for the blue line: I cannot find it in the editor. I had a crash of the editor when I tried to generate a silkscreen and then recovered from the autosave file, so maybe this line was drawn from the generate silkscreen function?

PR #308, opened 2023-02-01
Add Texas Instruments ADS101x ADCs

drakedevel:
Adds the TI ADS101x family of 12-bit delta-sigma ADCs in the VSSOP (DGS) package.These share a datasheet: https://www.ti.com/product/ADS1015

The footprint and symbol were exported from TI in KiCad format, imported, and touched up to pass checks. The 3D model was exported from TI as a STEP file, imported, and manually aligned.

I haven't made a board with this yet, but one is in the works, so I should be able to report back with test results once that's done.

### DIFF
--- a/entities/ic/adc/ti/ADS101x.json
+++ b/entities/ic/adc/ti/ADS101x.json
@@ -1,0 +1,19 @@
+{
+    "gates": {
+        "fabe58f7-ba88-4dfa-a17c-9932387834fe": {
+            "name": "Main",
+            "suffix": "",
+            "swap_group": 0,
+            "unit": "a3d29515-8422-4d40-83d1-fbd099c34d1e"
+        }
+    },
+    "manufacturer": "Texas Instruments",
+    "name": "ADS101x",
+    "prefix": "U",
+    "tags": [
+        "adc",
+        "ic"
+    ],
+    "type": "entity",
+    "uuid": "6589d0e1-ae31-4a5b-97e0-996599c83182"
+}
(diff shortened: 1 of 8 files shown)


carrotIndustries:
Bot!

drakedevel:
Update silkscreen on footprint for consistency with other parts, and add pin 1 indicator.

drakedevel:
Aha, it looks like the Pool manager omitted the STEP file from the PR. I couldn't figure out how to add it from the UI so I just added it manually.

drakedevel:
Shrank the symbol to a more reasonable size.

RX14:
> Aha, it looks like the Pool manager omitted the STEP file from the PR.

Could you submit this as a bug in the horizon-eda/horizon repo?

RX14:
Bug!

RX14:
Bot!

RX14:
I think this package might better be called VSSOP-10 (3x3mm) than TI's manufacturer specific code. The pad size and X pitch seem to be somewhat different than TI's own drawing, which specifies round metric 0.3x1.45 pads, and 4.4mm X pitch. It might also be worth using the rounded corner schematic pad with a corner radius of 0.07mm (1/4 the shortest side, rounded to 0.1mm).

For schematic symbols in horizon, it's good practice to arrange pins by function and not footprint layout. The power pin should be at the top and ground at the bottom. See for example TI's `DAC081S101` for a relevent example of a good horizon symbol, especially note the REFDES and VALUE placement. I would put all the AIN pins on the left and the data out on the right, for a left to right flow of information in the resulting schematic.

It's also ideal that there are different symbols for the ADS1015 and other parts, to stop errors where the wrong part is chosen and inputs are attached to the NC pins. This would also necessitate an alternate base part/unit/entity. If you are using only one of the two types, don't feel the need to complete the other.

Thanks so much for your first contribution, and apologies for all the nitpics.

PR #310, opened 2023-02-19
add stm32f105RxTx MCU made of separate blocks

LHSmicius:
Tried to make this part exactly as described in pool documentation. Just say if I misunderstood something or I can make something better.
Parts are made as common units, so these units can be reused to make other parts.
image


carrotIndustries:
Bot!

carrotIndustries:
This part could use the alt names for alternate pin functions.

LHSmicius:
Bot!
Thank you for notice. One more place to look at is symbol naming. I made symbols as reusable units that can be used in many MCU entities, but problem is that symbol can be assigned to one unit only, so it's not much sense to name them as I named "MCU GPIO port A 0-15". Maybe I should rename them to "STM32F105 port A", but they look very generic....
image

RX14:
> but they look very generic....

I think having these for each MCU is fine, they're very easy to duplicate and modify the alternate pin functions. It's easier than introducing something to capture this complexity in the pool data model. So these should probably be renamed to not be generic.

LHSmicius:
Bot!
I see, for now it's better to have symbols for each unit. That's fine :)

LHSmicius:
Bot!

RX14:
Thanks!

The power unit should be renamed to "STM32F105 Power" and the Vss/Vdd pins should be combined into one since they are the same net internally. And a tiny thing: `Boot_0` is referred to as `BOOT0` in the documentation i can see, and would make it consistent with `BOOT1`.

I think that's all till this can be merged

LHSmicius:
Bot!

LHSmicius:
Bot!
Hope this is the right way of declaring multiple Vdd & Vss pins of same net. In Part Editor I used same Vdd/Vss for multiple pins.

RX14:
This looks good to me. @carrotIndustries ?

tkornack:
Minor, disregardable opinion: 
- It's a different organization relative to all the other stm32 in the library, which use a single large symbol.
- I don't love the sideways power symbol. I would tend to put the power nets on Main symbol.

LHSmicius:
I think there can be different naming for parts made of multiple units, like STM32F105RBT6_sep (for separated). Maybe other word would fit better. _exp (for exploded, like exploded view in many CAD programs). Just brainstorming. This would allow to have multiple versions of same part.

LHSmicius:
As far as I can google the word "split" most often used for describing this part splitting into logical units. I can open an issue for a discussion on how to separate these two versions of the same part in library so that both versions would exist and everybody could chose his preferred version.

RX14:
I think the F105 is large enough to use multiple symbols.

LHSmicius:
I agree. I would never use large symbols, because I think it's a good practice to split schematic to logical blocks and avoid global nets as much as possible.

carrotIndustries:
LGTM, though I think we don't need the alt names on the power pins.

LHSmicius:
Bot!
image
If we agree on issue #312 that splitting unit is needed for MCUs when pin count is 100 and above. If we not agree I can reset to previous commit, no problem. But there are other 64 pin MCUs in the pool made in single unit, so perhaps it makes sense to not change them all.

tkornack:
Is it wide enough to accommodate typical alternate pin labels?

carrotIndustries:
> Is it wide enough to accommodate typical alternate pin labels?

Good point, this symbol should be expandable.

LHSmicius:
Can anybody explain what it means expandable? Any example from pool what to do?

LHSmicius:
Bot!
image
Nice feature!

PR #315, opened 2023-03-24
Ap62XXXWU step down converters

LHSmicius:
All AP62XXXWU voltage converters added.
Also TSOP-26 package added.
image


LHSmicius:
Bot!
I didn't add .step file for 3D view, but I would like to make it, just point me how to do it, please.

LHSmicius:
Bot!

LHSmicius:
Bot!

LHSmicius:
Bot!

RX14:
I think sot-23-6 is identical to (t)sot-26. I think these parts should use sot-23-6 package but perhaps sot-23-6 should be renamed to include a reference to sot-26. Everything else looks great!

LHSmicius:
SOT-23-6 just a little bit wider (0.2 mm). Crazy :DD
SOT-23-6:
image
TSOT-26:
image

RX14:
I think that difference is as much down to manufacturer variance than anything else, and in practice the same footprint will always work for both.

LHSmicius:
Bot!

LHSmicius:
Bot!

PR #316, opened 2023-03-26
Fix all mosfets in lib

LHSmicius:
A lot of work....

### DIFF
--- a/entities/transistor/mosfet-n-diode.json
+++ b/entities/transistor/mosfet-n-diode.json
@@ -1,19 +1,19 @@
 {
     "gates": {
-        "6dce85e8-6b70-44ac-8ae8-d17152c3707a": {
+        "25a1290b-4b6c-4cd3-b5b3-eac32ec98a90": {
             "name": "Main",
             "suffix": "",
             "swap_group": 0,
-            "unit": "375ee5ff-c8af-4b5e-bf2f-a59f8eb54fc3"
+            "unit": "ba993bf2-5791-44a5-854a-94862bb6ca39"
         }
     },
     "manufacturer": "",
-    "name": "P-channel MOSFET",
+    "name": "N-channel MOSFET with diode",
     "prefix": "Q",
     "tags": [
         "mosfet",
         "transistor"
     ],
     "type": "entity",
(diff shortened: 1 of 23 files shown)


LHSmicius:
Bot!

RX14:
Could you give an overview of the changes made in this PR so that this can be reviewed more easily?

LHSmicius:
I'm glad you asked :)
1. Made symbol "N-channel MOSFET with diode". Also unit and entity for it.
image
It better describes mosfets IMO and this diode is often integrated (long time I didn't see mosfet without flyback diode, actually).
Some mosfets have these gate over voltage protection diode, but I decided to not draw them (they doesn't effect schematic if voltage is in range).
image

2. Made base parts of N and P type mosfets with different packages:
- base N-MOSFET wDiode SOT-523 GSD
- base P-MOSFET wDiode SOT-23 GSD
- base P-MOSFET wDiode SOT-523 GSD
GSD means pin order Gate Source Drain.
Actually I was thinking if these BASE parts helps much, because derived parts inherits only Pin map, tags and package, but for low pin count packages it makes sense, because these packages are very popular.

3. Remake some mosfets (in order to change entity to version with diode).

4. Change part's MPN namings where orderable MPNs were combined with MPN. Also base mosfets renamed for better ordering (when ordering by MPN).

5. Change descriptions to standardize them. V, A, Ω.

6. Add missing tags.

RX14:
I'm concerned that with the new MOSFET entity a lot of existing schematics will break. If all the existing MOSFETS have diodes, then perhaps the existing entity should be renamed so that it has the same ID and existing schematics continue to work. This way there is no unit/symbol around with no (non-base) parts. The non-diode symbol can be kept as an alternative to the one with the diode.

LHSmicius:
Good point. So I will make another PR from master, because this one is too messed up. I'm planning to edit existing symbols "N-channel MOSFET" and "P-channel MOSFET" so that this version will have diode (by default all mosfets do). Only I will have to change entity for last PR dc29ff1c3d032184e4ebc22b31f87c5c9c82e1a2 , most likely it was not used in any schematic (week ago it was merged)

LHSmicius:
This PR can be closed, because PR https://github.com/horizon-eda/horizon-pool/pull/318 is a better version of these changes.

PR #319, opened 2023-04-05
Fix 2 pin JST XH connector package

frmdstryr:
Sorry I should have caught this earlier. The 2 pin package was not centered so rotating it shifted it around. It also was missing the package outline.


frmdstryr:
Bot!

frmdstryr:
Bot!

PR #320, opened 2023-05-14
add mosfet DMP610DL

LHSmicius:
Bot!
As always I have a question.
Is it right to inherit from other similar part which is not a base part?
Here I made DMP610DL mosfet which inherits from DMP3099L. These have much in common except electrical characteristics.
image

### DIFF
--- a/parts/transistor/pmos/DMP3099L.json
+++ b/parts/transistor/pmos/DMP3099L.json
@@ -3,6 +3,7 @@
         false,
         "DMP3099L"
     ],
+    "base": "03b1c077-db19-40a1-9a65-208053aff665",
     "datasheet": [
         false,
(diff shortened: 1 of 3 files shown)


LHSmicius:
Bot!

RX14:
> Is it right to inherit from other similar part which is not a base part

You should create a new base part for "sot-23 p-channel mosfet" and then set that as the base part for both parts. Then it matches up with how the rest of the pool is done.

LHSmicius:
Bot!

Made base part "base sot-23 P-channel MOSFET"
Changed base part for DMP610DL
Add base part for DMP3099L

RX14:
Looks good to me, thanks for the contribution!

PR #321, opened 2023-06-12
add bidir esd/tvs diode

LHSmicius:
As always I have some questions :D
**Unit**
I names pins "1" and "2", because it's bidirectional - it would be not accurate to call pins "A" and "K".
**Symbol**
I marked pin 1 with "+" because even if it's bidirectional (from electrical view), ESD and TVS diodes often has not symmetric package. Maybe pad numbers are enough... IDK...
image
**Package**
Just noticed that there is not described in "Horizon Pool Convention" that Courtyard expansion should be 0.25mm. It would be nice to mention somewhere.
image
**Part**
Base part has no description (It doesn't make any sense), so bot will fail. I think we need exception for base parts.


LHSmicius:
Bot!

LHSmicius:
!Bot
Finally text placement is correct. Bot fails on base part which is good.

LHSmicius:
Bot!

RX14:
If in doubt, look at what other parts do! The other diodes in the pool all use A/K pin naming.

In regards to the + pin marking, I am sceptical but I haven't used any of these diodes so I trust you to choose a sensible course. Other passives don't tend to enable pin markings on their symbols.

I would consider rounding the pads of the SOD package.

LHSmicius:
Changed pin names to A/K - you are right, even it's bidirectional, it makes more sense then naming 1/2.
Decided not to use "+" marking because it's non-standard and maybe it will bring more uncertainty.
Now I also think about round pads, it's easy to do, but then it will differs from other packages in the pool. I can make other PR and change all SOD packages to use round pads if needed

RX14:
Newer parts to the pool all have round pads. Square or round pads is purely about the age of the part in the pool, nothing else.

LHSmicius:
Didn't find how to calculate corner radius, so for now used default 0.1 mm. If it's not correct, I will fix it.

PR #323, opened 2023-07-29
add analog 8ch mux/demux

LHSmicius:
As always bot will fail because base part doesn't have description.
I'm not sure about multiplexer's I/O naming. I made base part according Nexperia datasheet where channels named Y0, Y1,..., Y7, and output (let's call it so) named Z. On internet I found many examples with different namings. Our logic components (AND, OR, XOR.....) already in pool uses letters A, B, C, .... for channels and 'Y' for output. Maybe I should continue that way.... Every manufacturer uses their own naming for multiplexer's I/Os...

### DIFF
--- a/entities/logic/single-mux-demux-8ch-en.json
+++ b/entities/logic/single-mux-demux-8ch-en.json
@@ -1,0 +1,24 @@
+{
+    "gates": {
+        "7b5a55c1-0b5c-4833-8d9a-82fbf512ed6f": {
+            "name": "Power",
+            "suffix": "P",
+            "swap_group": 0,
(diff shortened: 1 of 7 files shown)


LHSmicius:
Bot!

image

RX14:
The dual power is a bit funky, @carrotIndustries any ideas?

LHSmicius:
Bot!
I don't like that every manufacturer uses different pin names for such a basic device as multiplexer, so I thought it would be nice to have same naming in all logic components in the pool, but maybe I am wrong.... Maybe it should be according to datasheet...
image

PR #325, opened 2023-08-23
add STM32F103CxTx

LHSmicius:
image

Not sure:
- Which term is better to use "KB" or "KiB" for Kbytes in part description.
- Do I need to repeat "MPN" in "orderable MPNs" if it's the same? STM32F103CBT6 in this case.

This duplicates with https://github.com/horizon-eda/horizon-pool/pull/11.

### DIFF
--- a/entities/ic/mcu/stm/STM32F103CxTx.json
+++ b/entities/ic/mcu/stm/STM32F103CxTx.json
@@ -1,0 +1,22 @@
+{
+    "gates": {
+        "e735a092-a6d7-4498-be6e-36d9707b9f5e": {
+            "name": "Main",
+            "suffix": "",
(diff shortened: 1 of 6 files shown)


LHSmicius:
Bot!

auroraanna:
> Which term is better to use "KB" or "KiB" for Kbytes in part description.

i'd go for "KiB" if you mean 1024 byte. else, "KB" is just wrong :/

tkornack:
I'm torn about whether to use alternate names for PC13-PC14 and PA0's alternate functions. You've done it just like the data sheet, and that's the horizon standard, so I suppose it's good. I also prefer the capitalized VSS/VDD. Otherwise looks good to me.

PR #329, opened 2023-09-15
Add ABL-8.000MHZ-B4Y

ii8:


ii8:
Bot!

ii8:
There is actually a 0.2mm gap between the package and the silkscreen, I believe the clearance check isn't working right because of the rounded part of the package.

RX14:
Hi, could you confirm the source of the step file and the licence terms it is under? If the licence is incompatible, perhaps there is an equivalent model from kicad's 3d model collection

ii8:
The step file is from Abracon, I couldn't find license information on their website and I've emailed them but still no response after several days.

PR #330, opened 2023-09-19
add xCA9555PW I2C I/O expander

LHSmicius:
Added these parts:
xCA9555PW_Base
+-- PCA9555PW
+-- TCA9555PW
+-- NCA9555PW
+---- NCA9555PW-Q100 (same as NCA9555PW, just automotive certified)


LHSmicius:
Bot!

LHSmicius:
Bot!
Made separate units for NXP, Nexperia and Texas Instruments xCA9555 chip. It's nice because pin names are exactly as in datasheet, but it uses a lot more storage to store all these very similar parts.

PR #335, opened 2023-11-02
add STM32H730VBT6 mcu

LHSmicius:
add STM32H730VBT6 mcu and needed package QFP100


LHSmicius:
Bot!

tkornack:
Was there a consensus to use split symbols for these larger parts? I'm working on H743 that uses split symbols.

LHSmicius:
Bot!
> Was there a consensus to use split symbols for these larger parts? I'm working on H743 that uses split symbols.

In convention it is written "if they have more than 100 pins"

LHSmicius:
Bot!

LHSmicius:
Bot!

PR #341, opened 2023-12-06
add zener, fix some other diodes

LHSmicius:
Add zener symbol and base zener part on SOT-23. Created only base zener diode, because there are so many of them (including those zener voltage differences), so I use base parts in my projects and change "Custom value" field. 
Reference: https://assets.nexperia.com/documents/data-sheet/BZX84_SER.pdf

Also fix some minor mistakes on other diodes.


LHSmicius:
Bot!

RX14:
IRC 60617 shows 

might be good to change the symbol to match that. Also keep in mind that your text placement isn't quite right for all the different orientations.

LHSmicius:
Bot! 
Nice to see you back, Stephanie!

PR #343, opened 2024-02-12
LED lib changes

LHSmicius:
Add LED 1411 (3528 metric) package with one part and fix some other LED parts.
Metric package name isn't written anywhere. Should I add it?


LHSmicius:
Bot!

PR #346, opened 2025-08-10
Add ESP32-C6 MCU

jue89:
* package: generic QFN-40
* symbol, unit, entity: ESP32-C6
* part: ESP32-C6


jue89:
Bot!

RX14:
Thanks for the contribution! Usually for chips this small we don't split them into separate units. Especially here where there's an io mux that lets you reassign the pins anyway. USB_D+ and USB_D- also seem to have their GPIO numbers wrong in the alternate functions, might be worth double-checking the mapping since I didn't check them all.

jue89:
Thanks for your feedback :) I'll address it during the days.

jue89:
Bot!

jue89:
I really have no idea how time can run this fast ... but here we go. Thank you again four your feedback, @RX14!

I merged everything into one unit/symbol/entity, as you requested.

The idea behind the grouping of pin functions:
- CHIP_PU, ANT and XTAL_* are analog pins with no other function
- USB_* are the default function of that pins and almost every project I've seen on the internet using this MCU is connecting an USB port to those pins
- VDD_SPI is a power output for an external NOR flash, SPI* are the interface pins for the external flash storing the application. This MCU is close to useless without any external flash. GPIO26+28 are handshaking lines for the external flash.

All the other pins on the right side of the symbol can be perfectly used are GPIOs, despite them being muxed to peripherals after reset.

Also thank you for pointing out the confused USB interface :) You saved me two botch wires on the first PCBs based on this MCU ;)

jue89:
Bot!

And now including the part ...

RX14:
If you've tested this package and symbol on a real board, then it seems ready to go to me. I'll trust you on the GPIO/alternates assignments.

The only note I have is that the GND pin can probably have it's text rotated 90 degrees

jue89:
Bot!

PR #347, opened 2026-01-17
Add TSL25911 Ambient Light Sensor

B48D81EFCC:
**ams TSL25911 Ambient Light Sensor**

- I2C sensor
- also new package added: ODFN-6

There is still one warning: **Use polygons to draw outlines on assembly layers.**
Can someone help me to fix this issue?


B48D81EFCC:
Bot!

B48D81EFCC:
Fixed a filename typo

tkornack:
It looks like the symbol is missing a box.

Also seems as though the package needs work:
- the footprint pad sizes do not follow the manufacturer's recommendations. 
- the assembly layer should show a rectangle with the dimensions of the part
- The reference designator is too small and looks like it's at an odd angle?

RX14:
There's two draw rectangle tools: draw line rectangle and draw polygon rectangle. I think you may have used the wrong one on the assembly layer.

B48D81EFCC:
Bot!

tkornack:
Improved! 
- Fix the whitespace in the symbol name
- The pin 1 dot should be removed from the package layer
- The reference designator may be rotated 90 and expanded to fill the package with the M1234 text.
- Symbol should be schematic, not a physical map. Power pins should be on the top and bottom of the symbol box. Use the ‘perpendicular’ name orientation unless this forces the symbol to be unreasonably wide.
- A pin 1 indicator is established by extension of an existing outline or by omitting a line. Do not use a dot or text or any other marking.

B48D81EFCC:
I think it look beautiful now!

B48D81EFCC:
Bot!

B48D81EFCC:
Bot!

tkornack:
This is what I meant about the power pins:
image
And this is an example from UFQFPN32 to see how the silkscreen can be done: 
image
Remove the pin 1 circle.
The package layer should be the package outline only
The assembly layer should have the pin 1 chamfer and the reference designator.

B48D81EFCC:
Fingers crossed

B48D81EFCC:
Bot!

B48D81EFCC:
Bot!

RX14:
Has this been succesfully assembled on a board?

B48D81EFCC:
> Has this been succesfully assembled on a board?

No, not yet

RX14:
Once that's complete I'm OK to merge

tkornack:
ignorable nit: in the symbol the NC pin probably shouldn't be the first one on top

RX14:
the NC probably shouldn't be there at all, unlike kicad we don't usually put NC pins in the symbol

tkornack:
Yes, in this case it's clearly "do not connect", so I agree it should be removed from the symbol. (Sometimes with "no internal connection" you can have a reason to connect it to ground to have a more complete ground plane.)

PR #348, opened 2026-01-22
Add BME280 (Bosch Sensortech)

B48D81EFCC:
Includes BME280 sensor ic


RX14:
Bot!

RX14:
Your assembly outline should be the same size as your package rectangle except with a pin 1 marker, for typical packages. And the courtyard tool should be run last. You also seem to be missing the rectangle on your symbol

B48D81EFCC:
- Fixed assembly outlines
- Fixed symbol

B48D81EFCC:
Bot!

RX14:
Bot!

RX14:
Have you seen the documentation [here](https://docs.horizon-eda.org/en/latest/create-package.html)? It's worth checking that and opening a few existing packages to see how they work and look. 

The courtyard is good, and the assembly outline is much better, but the assembly and silkscreen is still missing a pin 1 marker. While drawing the assembly rectangle you will see the decoration tools for creating that marker in the bottom. Check the two existing DFN packages for how the pin 1 markers on those packages work.

B48D81EFCC:
Bot!

B48D81EFCC:
@RX14 Thanks for the guidance and patience.
Should be much better now.

tkornack:
Please consider:
- Replace the pin 1 dot with a chamfer
- Study how silkscreen pin 1 mark is done for similar headless packages
- In the symbol: the two ground pins should probably be a single ground pin
- The power pins are preferred on top and bottom of the symbol, with horizontal text.

PR #350, opened 2026-04-12
Nichicon UES series capacitors

atoav:
This adds the Nichicon UES series capacitors (bipolar electrolytic capacitors often use for analog audio inputs and outputs).

I generated the parts using scripts in the scripts directory


RX14:
Is this a candidate for another pool? I think there was a rough consensus to keep big new generated series in their own pool for performance reasons.

PR #351, opened 2026-06-05
add SQFN-36 package and USB251xB USB hubs

auroraanna:
USB251xB is a series of USB 2.0 USB hubs with either 2, 3 or 4 downstream USB ports which use the SQFN-36 package.

### DIFF
--- a/entities/ic/interface/usb/microchip/USB2512B.json
+++ b/entities/ic/interface/usb/microchip/USB2512B.json
@@ -1,0 +1,38 @@
+{
+    "gates": {
+        "585f3867-3b4a-4827-8bca-d600768cc1fc": {
+            "name": "NC",
+            "suffix": "D",
+            "swap_group": 0,
+            "unit": "fc837648-ee97-47d0-87ea-6a40c89f6fd7"
+        },
+        "6ec5e3b0-b175-4fc3-a5a2-a358d2f11cad": {
+            "name": "common",
+            "suffix": "A",
+            "swap_group": 0,
+            "unit": "1c458e78-ade5-48d0-9f3d-58751f387744"
+        },
+        "a7c8037c-c23e-492b-9031-332112ad5e09": {
+            "name": "downstream data 1",
+            "suffix": "B",
+            "swap_group": 0,
+            "unit": "05d613bb-1f3e-4722-a1ba-fa7181c28ddf"
+        },
+        "cd718c70-835d-4b24-b724-f90445fae99d": {
+            "name": "downstream data 2",
+            "suffix": "C",
+            "swap_group": 0,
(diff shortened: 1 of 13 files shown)


auroraanna:
SQFN-36 does seem to be used exclusively by Microchip Technology after all.

auroraanna:
For the remaining rules checks not passing, I don't know what those mean or how to fix them.

On the SQFN-36 package, the check "Use polygons to draw outline on assembly layers" has result "Warn" even though I do have an outline of the package drawn on the top assembly layer.

On the USB251xB common symbol there's a bunch of errors about pads not being in the bounding box but I don't know what the bounding box is or how I can modify it.

auroraanna:
Bot!

auroraanna:
Changed XTALIN/XTALOUT pins in USB251xB common symbol not to be clock since a crystal is also allowed and moved GND pin to the bottom where there's more space.

auroraanna:
…except for the numbers, since that's not possible with multiple gates.

auroraanna:
I don't know how to squash these commits now because of the weird branch that the program made :/
I think I made all the fixes I now though I forgot when publishing the PR.

auroraanna:
There is a STEP file available for the package on DigiKey but the license does not allow redistribution outside of your organization. The model also has no pin 1 indicator.

auroraanna:
oh no, i placed a bunch of files in wrong locations…

auroraanna:
figured out that that branch is on my fork, moved the files to better locations and squashed commits.

tkornack:
Generally for smaller pin count devices you shouldn't be splitting the symbol. See https://github.com/horizon-eda/horizon-pool-convention . For the NC pin you should either place it on the main symbol or leave it out of the symbol completely. The symbol needs to be symmetric around the origin so that the pin tips sit on the bounding box.

auroraanna:
Integrating the NC pin into the common symbol is not an option because USB2514B does not have a NC pin. Leaving it out entirely would make sense I guess since nothing should be connected there from the side of the IC anyway…

I think it makes sense to have multiple units due to the multiple parts that have are the same except for the downstream port gates and the NC pin.

RX14:
Bot!

RX14:
Could you check the warnings on the symbol? it seems your symbol might not be symmetrical around the origin

auroraanna:
fixed!

tkornack:
Have you designed with this part? The XTALOUT/IN are on opposite sides of the part. 

Also, the ground pin should use a horizontal label (and likely be centered). I would also strive to make the supply pins be symmetric.

I think the NC should not be a separate unit/entity. I suggest removing it entirely or placing it on the main symbol 

Finally, I still believe that this should be a single symbol.

auroraanna:
ok i'm gonna make it into one symbol :~:

PR #352, opened 2026-07-10
add various CRCW0603 resistors

griush:
Adds 4 Vishay resistors that I'm using on a project.


RX14:
Bot!

RX14:
Could you create a base package for CRCW0603 similar to the CRCW0402 base which is already in the pool

griush:
Thanks for the comment, please check if that is better.

tkornack:
Bot!

